PR #44, opened 2019-04-14
Ajouter une resource -> Add resource

tenrol:
Traduction de francais en anglais de "Ajouter une resource" dans le container mapping

### DIFF
--- a/client/src/application/containers/utils/navbar/index.tsx
+++ b/client/src/application/containers/utils/navbar/index.tsx
@@ -1,3 +1,4 @@
+
 import {
     Alignment,
     Button,
@@ -8,6 +9,7 @@ import {
     Toaster,
 } from '@blueprintjs/core'
 import * as React from 'react'
+import { Redirect } from 'react-router-dom'
 import {
     Mutation,
     Query,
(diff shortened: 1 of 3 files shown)


client/src/application/containers/utils/navbar/index.tsx
```diff
@@ -112,7 +129,8 @@ class Navbar extends React.Component<IProps, IState> {
 
         const logo = <BPNavbar.Heading>
             <span dangerouslySetInnerHTML={{__html: arkhnLogoWhite}} />
-            <h2>PYROG</h2>
+            {this.renderRedirect()}
+            <h2 onClick={this.setRedirect}>PYROG</h2>
```
alexis-thual: Just a few comments:
* Rather than ```
onClick={this.setRedirect}
```
try and use this:
```
onClick={this.props.history.push('/')
```
It will allow this redirection to be stored in React redirection history.
* Try to use `onClick` directly in `BPNavbar.Heading`

File: client/src/application/containers/utils/navbar/index.tsx [hunk 2]
```diff
@@ -77,6 +79,21 @@ class Navbar extends React.Component<IProps, IState> {
...
+        this.setState({
+            redirect: true
+        })
+    }
+    renderRedirect = () => {
+        if (this.state.redirect) {
+            return <Redirect to='/' />
+        }
+    }
+
```
alexis-thual: This part should become useless if you use `this.props.history.push` as mentioned in the other comment :)

PR #47, opened 2019-04-16
Feature/app state in url

alexis-thual:
In this PR:
* in the `mapping` view, the app state is now present in the URL
* prettify code


saifali96:
Swift work on on the FR! 💯 

Do you think it's better to move the prettified code to a separate branch and PR, to help in managing diff size and codebase tracking?

alexis-thual:
Haha, too late, I pushed this code, rewriting history is bad :slightly_smiling_face:
But you are absolutely right, it belonged in another PR and I should have done a different one.

PR #49, opened 2019-04-18
Circle ci

alexis-thual:
`circle ci ` integration to run tests.
Tests are now all launched from docker containers. A new container is setup in `docker-compose.test.yml` and launches a test environment (which is similar to the environment pyrog server runs on) based on our mock database.


server/docker-compose.test.yml
```diff
@@ -0,0 +1,50 @@
+version: "3"
+services:
+  test:
+    build:
+      context: .
+      dockerfile: Dockerfile.test
+    image: arkhn/pyrog_test:lastest
```
LaRiffle: isn't it "latest"

alexis-thual: Yes indeed my friend, it is now solved!

alexis-thual:
I think CI now works for `server`.
We should now implement tests for `client`!

Issue #40, opened 2019-04-06
Typo in MIMIC mappings rules 


LaRiffle:
As for now the mappings rules come back like this:
```
          "name": "birthDate",
          "mergingScript": null,
          "isProfile": null,
          "type": "date",
          "inputColumns": [
            {
              "id": "cju2uylrf02q70734a6103ooc",
              "owner": "MIMIC",
              "table": "PATIENTS ",
              "column": "DOB ",
              "script": null,
              "staticValue": null,
              "joins": []
            }
          ],
```

- There is a pb with the  space after the table and the column name
- There is no owner in the mimic scheme, so owner should be set at `null`

--------

Query used:
```
fragment entireJoin on Join {
        id
        sourceOwner
        sourceTable
        sourceColumn
        targetOwner
        targetTable
        targetColumn
    }
    
fragment entireInputColumn on InputColumn {
    id
    owner
    table
    column
    script
    staticValue
    joins {
        ...entireJoin
    }
}

fragment a on Attribute {
    id
    comment
    name
    mergingScript
    isProfile
    type
    inputColumns {
        ...entireInputColumn
    }
}
    
query {
  resource(where: {id: "cjtl9lxn400hk0934ax2qy8nm"}) {
    id
    name
    attributes {
          ...a
          attributes {
                    ...a
                    attributes {
                        ...a
                        attributes {
                            ...a
                            attributes {
                                ...a
                                attributes {
                                    ...a
                                    attributes {
                                        ...a
                                        attributes {
                                            ...a
                                            attributes {
                                                ...a
                                                attributes {
                                                    ...a
                                                    attributes {
                                                        ...a
                                                        attributes {
                                                            ...a
                                                            attributes {
                                                                ...a
                                                                attributes {
                                                                    ...a
                                                                    attributes {
                                                                        ...a
                                                                    }
                                                                }
                                                            }
                                                        }
                                                    }
                                                }
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
    }
  }
}
```

alexis-thual:
It kinda hacky but I just updated the mockdata. You should be able to have this solved if you redownload them and re-setup the database. Let me know!

PR #50, opened 2019-04-22
Update node modules

alexis-thual:
Two simple node module updates:
* `react` and `react-dom` for `client`
* `prisma` and `typescript` for `server`

People might have to rm and reinstall node_modules.


alexis-thual:
Since our datamodel needed some refactoring (adding `"link": TABLE` to `@relation` so as to indicate that join should be performed using a third table), I also modified our mock data to fix #40 and remove irrelevant `OWNER`.
We can now specify `hasOwner` in a `Source`.
**As this is a breaking change, one should redeploy prisma and repopulate it with freshly downloaded mock data.** @LaRiffle

PR #61, opened 2019-09-23
Multiple instances of the same fhir resource for a given source

### LINKED ISSUE #56: Allow multiple declarations of similar FHIR resources
In some cases, multiple mappings are possible for a given FHIR resource.

For example, we could map a `Observation` resource to store the `Clinical Sex` of a `Patient`, on a given column from the source database. Then map "another" `Observation` resource for `Blood Pressure` on rows from a specific table storing these values in the source database, etc.

For this purpose, we should : 

- [x] UI - Be able to add multiple resources from the same FHIR type
- [x] UI - Be able to "name" a mapped resource to differentiate it from the other resources of the same type (labelling)
- [x] DB - Store the actual `name` of resources in a `FHIR type` field, to have the `name` field dedicated to a "labelling" value
- [x] FHIRPIPE - Take this new feature into account for data migration

alexis-thual:
Closes #56 

Allows instantiating the same fhir resource several times for a given source.
See #56 for detailed feature list.


alexis-thual:
Tests will not pass for now as someone needs to update mock data with the new prisma datamodel.

Nabellaleen:
Replaced by #63

PR #66, opened 2019-10-02
Handle credentials when source is a database

alexis-thual:
* Updates all react related dependencies
* Refactors `Navbar` using react hooks
* Add `Credentials` to datamodel
* Allow upserting `Credentials` from client


client/src/components/Navbar/components/Drawer/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+import { useSelector } from "react-redux";
+import { useApolloClient } from "react-apollo-hooks";
+
+import "./style.less";
+import StringSelect from "../../../selects/stringSelect";
+import { IReduxStore } from "../../../../types";
+import sourceSchemas from "../../../../services/selectedNode/sourceSchemas/reducer";
+
+const credential = require("../../../../graphql/queries/credential.graphql");
+const upsertCredential = require("../../../../graphql/mutations/upsertCredential.graphql");
```
simonvadee: nit: I tend to use absolute paths in my imports as it improves readability. eg: `import StringSelect from 'components/selects/stringSelect';` I think it shuyold work out-of-the-box with `react-scripts`

client/src/components/Navbar/components/Drawer/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const [hasChanged, setHasChanged] = React.useState(false);
+  const [hasSuccessfullyChanged, setHasSuccessfullyChanged] = React.useState(
+    false
+  );
+  const [loading, setLoading] = React.useState(true);
+  const selectedNode = useSelector((state: IReduxStore) => state.selectedNode);
+  const client = useApolloClient();
+
+  React.useEffect(() => {
+    console.log("USE EFFECT", selectedNode);
```
simonvadee: nit: remove console.log

client/src/components/Navbar/components/Drawer/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+                        sourceId: selectedNode.source.id
+                      }
+                    });
+                  }}
+                >
+                  <FormGroup label="Host">
+                    <InputGroup
+                      value={host}
+                      disabled={loading}
+                      leftIcon={"desktop"}
```
simonvadee: where do these icons come from ? don't you have a react library where you can import them from ?

client/src/app.tsx
```diff
@@ -143,7 +144,9 @@ const token = localStorage.getItem(process.env.AUTH_TOKEN);
 ReactDOM.render(
   <Provider store={store}>
     <ApolloProvider client={client as any}>
-      <Routes />
+      <ApolloHooksProvider client={client as any}>
```
simonvadee: why do you need a `as any` here ?

server/src/resolvers/Mutation/pyrog.ts
```diff
@@ -227,5 +228,41 @@ export const pyrogMutation = {
...
+    parent,
+    { login, password, host, port, type, database, sourceId },
+    context: Context
+  ) {
+    getUserId(context);
+    const encryptedPassword = await bcrypt.hash(password, 10);
+    const credential = await context.client
+      .source({ id: sourceId })
+      .credential();
+    console.log(credential);
```
simonvadee: nit: remove console.log

server/src/resolvers/Mutation/pyrog.ts
```diff
@@ -1,3 +1,4 @@
+import * as bcrypt from "bcryptjs";
```
simonvadee: only import the symbol you need instead of *

client/src/components/Navbar/components/Drawer/index.tsx
alexis-thual: They come from BlueprintJs, which ships with a set of icons!

server/src/resolvers/Mutation/pyrog.ts
alexis-thual: Fixed in cf867df

server/src/resolvers/Mutation/pyrog.ts
alexis-thual: Oops, my bad. Useless `console.log` are removed in 1460525

client/src/components/Navbar/components/Drawer/index.tsx
alexis-thual: Great, I didn't know one could do this! Fixed in 8f3a6c1.
I'll slowly refactor other components and start using absolute paths instead.

client/src/app.tsx
alexis-thual: Hummm, that's a nice pick. I previously use `client as any` for `ApolloProvider` because of a mismatch in types between `ApolloProvider` and `ApolloClient` and simply assumed it would be the same for `ApolloHooksProvider`.

Actually, everything seems to be in order now, and I removed both "uncasts" in 5e6b787

Issue #23, opened 2019-02-15
How do we manage scripts


LaRiffle:
We should have a registry of scripts (maybe a python file / class / package), each of each would have a attr name and short description to help choice
We should have a way to add from the fhirball interface a "request" when a script is missing (maybe simply a link to the gihub issue panel and instructions to fill a clear issue). (I don't believe on python code writing from the fhirball interface)

PR #74, opened 2019-10-11
#23 Make the cleaning script select load info from cleaning_script

Nabellaleen:
# TODO
- [ ] Make the cleaning-script server URI configurable through env variables

---

Need https://github.com/arkhn/cleaning-scripts/pull/2


client/src/app.tsx
```diff
@@ -126,14 +127,25 @@ const link = split(
   wsLink,
   httpLinkAuth
 );
+const links = [];
+if (process.env.NODE_ENV === "development") {
+  links.push(errorLink);
+}
+links.push(
+  new RestLink({
+    uri: "http://127.0.0.1:5000/",
```
Nabellaleen: Here !
- [ ] Replace by an environment variable to make this URI configurable

PR #75, opened 2019-10-23
#23 - Make the cleaning script great again

Nabellaleen:
Build a select component querying the cleaning script flask server as suggested in https://github.com/arkhn/pyrog/issues/23

image


client/tsconfig.json
```diff
@@ -8,7 +8,10 @@
     "target": "es6",
     "allowJs": true,
     "lib": ["es6", "dom", "esnext"],
-    "baseUrl": "."
+    "baseUrl": ".",
+    "paths": {
+      "~/*": ["src/*"]
```
LaRiffle: What does does this one do?

client/src/components/selects/scriptSelect.tsx
```diff
@@ -0,0 +1,96 @@
...
+  let items: IScript[] = [];
+  if (data && data.cleaningScripts) {
+    items = data.cleaningScripts.scripts;
+  }
+  return (
+    <ButtonGroup>
+      <TSelect<IScript>
+        disabled={false}
+        displayItem={(item: any) => {
+          return item.code || "None";
```
LaRiffle: Can item.code be empty?

client/src/components/selects/scriptSelect.tsx
```diff
@@ -0,0 +1,96 @@
...
+  onClear
+}: IProps) => {
+  const { loading: queryLoading, error, data } = useQuery(cleaningScripts);
+  let items: IScript[] = [];
+  if (data && data.cleaningScripts) {
+    items = data.cleaningScripts.scripts;
+  }
+  return (
+    <ButtonGroup>
+      <TSelect<IScript>
```
LaRiffle: Stupid question, what is the semantic of `<TSelect<IScript> ... />` ?

client/tsconfig.json
Nabellaleen: - [ ] Ajouter un commentaire

Nabellaleen: Thank you ! I've added a comment :)

client/src/components/selects/scriptSelect.tsx
Nabellaleen: I'm not sure to understand what you're talking about semantic ;) Tell me if I don't answer to your question !

`TSelect<IScript>` means that we are implementing a `TSelect` taking as items only objects implementing the `IScript` interface.

LaRiffle: So it would be a `TSelect` but extended with attributes from `IScript`?

PR #80, opened 2019-11-19
Add Dockerfile for pyrog client

LaRiffle:
The different Dockerfile & docker-compose work correctly but they are not production ready. :/

Additionally, I have the feeling that pyrog-server is not waiting for prisma to be up and ready --> this causes errors the first time.


server/Dockerfile
```diff
@@ -1,44 +1,13 @@
-###############################################################################
-# Step 1 : Builder image
-#
-FROM node:12.11-alpine AS builder
+FROM node:10.15-alpine
```
simonvadee: why did you rewrite this Dockerfile ? I don't think the the image should be built like this, it should only contain the "naked" server (`yarn run deploy` and `yarn run import` should be run from the outside). Maybe it was written like this before and you made a mistake while merging the master branch into this one ?

client/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:10.15-alpine
+WORKDIR /usr/src/app
+COPY package.json yarn.lock ./
+RUN yarn install
+COPY . .
```
simonvadee: did you think of adding a `.dockerignore` file in order not to copy everything into the image (including nodes_modules) ?

server/Dockerfile
LaRiffle: I agree this is not the good structure (at all!), the problem was that the previous Dockerfile was not working correctly (it missed the part with pyrog_dev_static.zip)
I couldn't manage to fix it properly.
The key thing is for me to have yarn running inside the container, as we don't want to install yarn ourselves on some machines. We have real examples where we see it's a tall order... ;)

client/Dockerfile
LaRiffle: I'll add it!

server/Dockerfile
simonvadee: ok I get what you mean but still, you can still run yarn inside the container (`docker exec <the-container> yarn run deploy`) and mount the pyrog_dev_static file as a volume instead of adding it directly into the image. we can talk about it monday together !

client/Dockerfile
simonvadee: :+1:

server/Dockerfile
LaRiffle: Yes we'll do this :)
Speaking of `pyrog_dev_static`, I think we'll get rid of it and do like you did for pyfhirstore & api (download of the official website and clean). This way we make sure we have the same schema across the products, which is crucial.
Then the question of attaching the schema to the image is still left opened, it's tighly link to how we want to handle versioning (of the pyrog software, of FHIR, of the docker image, etc)

PR #97, opened 2019-12-17
Make the merging script great again

Jasopaum:
Enable selection of merging scripts

---
This change is [Reviewable](https://reviewable.io/reviews/arkhn/pyrog/97)

### DIFF
--- a/client/src/views/Mapping/components/InputColumns/index.tsx
+++ b/client/src/views/Mapping/components/InputColumns/index.tsx
@@ -73,7 +73,7 @@ const InputColumns = ({ schema, selectedAttribute, source }: IProps) => {
                               <ScriptSelect
                                 loading={loading}
                                 selectedScript={attribute.mergingScript}
-                                onChange={(script: string) => {
+                                onChange={(script: string) => (
                                   updateAttribute({
                                     variables: {
                                       id: attribute.id,
@@ -81,9 +81,9 @@ const InputColumns = ({ schema, selectedAttribute, source }: IProps) => {
                                         mergingScript: script
                                       },
                                     }
-                                  });
-                                }}
-                                onClear={(): any => {
+                                  })
+                                )}
+                                onClear={(): any => (
                                   updateAttribute({
                                     variables: {
                                       id: attribute.id,
@@ -91,8 +91,8 @@ const InputColumns = ({ schema, selectedAttribute, source }: IProps) => {
                                         mergingScript: null
                                       },
                                     }
-                                  });
-                                }}
+                                  })
+                                )}
                               />
                             </div>
                           );
--- a/client/src/views/Mapping/components/InputColumns/index.tsx
+++ b/client/src/views/Mapping/components/InputColumns/index.tsx
@@ -1,18 +1,17 @@
 import { Tag, Spinner } from "@blueprintjs/core";
 import * as React from "react";
-import { Mutation, Query, Subscription } from "react-apollo";
+import { Mutation, Query } from "react-apollo";
 
-import { ISelectedSource } from "../../../../types";
+import { ISelectedSource } from "src/types";
 
 // COMPONENTS
-import StringSelect from "../../../../components/selects/stringSelect";
+import ScriptSelect from "src/components/selects/scriptSelect";
 import InputColumn from "./../InputColumn";
 
 // GRAPHQL
-const inputColumns = require("../../../../graphql/queries/inputColumns.graphql");
-const updateAttribute = require("../../../../graphql/mutations/updateAttribute.graphql");
-const subscribeAttribute = require("../../../../graphql/subscriptions/attribute.graphql");
-const subscribeInputColumn = require("../../../../graphql/subscriptions/inputColumn.graphql");
+const inputColumns = require("src/graphql/queries/inputColumns.graphql");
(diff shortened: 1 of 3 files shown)


client/src/views/Mapping/components/InputColumns/index.tsx
```diff
@@ -37,51 +36,30 @@ const InputColumns = ({ schema, selectedAttribute, source }: IProps) => {
         let inputColumns = data && data.inputColumns ? data.inputColumns : [];
 
         return selectedAttribute.id ? (
-          <Subscription
-            subscription={subscribeAttribute}
+          <Query
```
simonvadee: why do you use Query instead of Subscription now ?

client/src/views/Mapping/components/InputColumns/index.tsx
```diff
@@ -92,22 +70,26 @@ const InputColumns = ({ schema, selectedAttribute, source }: IProps) => {
...
-                                  attribute && attribute.mergingScript
-                                    ? attribute.mergingScript
-                                    : ""
-                                }
-                                items={["mergingScript.py"]}
+                              <ScriptSelect
                                 loading={loading}
-                                onChange={(e: string) => {
+                                selectedScript={attribute.mergingScript}
+                                onChange={(script: string) => {
```
simonvadee: I don't think you need to have a block function here: `onChange={ (script:string) => updateAttribute({...}) }`

client/src/views/Mapping/components/InputColumns/index.tsx
```diff
@@ -92,22 +70,26 @@ const InputColumns = ({ schema, selectedAttribute, source }: IProps) => {
...
+                                  updateAttribute({
+                                    variables: {
+                                      id: attribute.id,
+                                      data: {
+                                        mergingScript: script
+                                      },
+                                    }
+                                  });
+                                }}
+                                onClear={(): any => {
```
simonvadee: same as previous comment

server/src/resolvers/Mutation/pyrog.ts
```diff
@@ -111,7 +111,14 @@ export const pyrogMutation = {
 
   updateResource: checkAuth(forwardTo('binding')),
 
-  updateAttribute: checkAuth(forwardTo('binding')),
+  async updateAttribute(parent, { id, data }, context: Context) {
```
simonvadee: nit: this function does not need to be async (and you don't need to return `await context.client....`, you can just drop the await I think)

Jasopaum: You're right! In that case, I should also remove the `await`s for the other functions too, am I right?

client/src/views/Mapping/components/InputColumns/index.tsx
Jasopaum: I'm not sure why but the subscription was not working as it was supposed to: the `data` returned was always null.
Also, I don't understand why subscriptions are needed for Pyrog, I think queries can always do what we want and are easier to manage.

client/src/views/Mapping/components/InputColumns/index.tsx
Jasopaum: Not sure I follow. What would you write instead?

client/src/views/Mapping/components/InputColumns/index.tsx
simonvadee: Subscriptions are a nicer way to update the UI using websockets without having to re-query the same fragment but I agree that it it simpler to manage for now and doesn't cost much anyway.

client/src/views/Mapping/components/InputColumns/index.tsx
simonvadee: the `updateAttribute` function does not need to be in a block, it can be used as return value. it's not very important though.

server/src/resolvers/Mutation/pyrog.ts
simonvadee: yes you can, it's always good to clean the project when you notice something isn't right at one place.

Jasopaum: 👍

client/src/views/Mapping/components/InputColumns/index.tsx
Jasopaum: I think that queries are cached and this cache is automatically changed when performing updates on a single object (in some other cases, it can be modified by hand).
Shouldn't this be preferred over having a websocket connection? (That's what I'm doing in the v2 so if you think it isn't we will need to talk about it ^^)

simonvadee: subscriptions are tricky anyway, and using the apollo cache is required whether we use subscriptions or not. I think it's always best to start with queries, so keep going! we'll talk about it in more details when i'll get back

client/src/views/Mapping/components/InputColumns/index.tsx
```diff
@@ -73,26 +73,26 @@ const InputColumns = ({ schema, selectedAttribute, source }: IProps) => {
                               <ScriptSelect
                                 loading={loading}
                                 selectedScript={attribute.mergingScript}
-                                onChange={(script: string) => {
+                                onChange={(script: string) => (
```
simonvadee: you don't even need the parenthesis

Issue #100, opened 2020-01-03
Enhance the mapping experience


elsiehoffet-94:
A few things could be added to improve the user experience and the mapping: 

- [ ] Add a visual indicator to highlight **required** attributes (which must be filled by the mapper). Required attributes are indicated by ** R!** in the JSON template of a resource, for instance in MedicationRequest.status:

`{
  "resourceType" : "MedicationRequest", [...] 
  "status" : "<code>", // R!  active | on-hold | cancelled [...] }`

- [ ] Add a comments/remarks section for each attribute. This could be used by the mapper to explain some decisions he or she has to take (for example the choice of a static value). This would not be part of the mapping rules.

PR #101, opened 2020-01-06
Enhance mapping experience

Jasopaum:
Solving issue #100 

I added comments and isRequired to the attribute data model (backend).

On the front it looks like a orange dot next to attributes that are required (or that have required children):
image

And a text area where you can add comments for each attribute:
image

Feel free if you have any idea about how to make this prettier/more intuitive!


server-v2/prisma/migrations/lift.lock
```diff
@@ -3,4 +3,5 @@
 # lift lockfile v1
 # Read more about conflict resolution here: TODO
 
-20191220162753-bootstrap
\ No newline at end of file
+20200103230912-bootstrap
```
simonvadee: you should have only one line here since you modified the migration in place

PR #103, opened 2020-01-10
v2

simonvadee:


client/public/index.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
```
LaRiffle: What is this for?

.github/workflows/server-v2.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Cache node modules
+        uses: actions/cache@v1
+        with:
+          path: server-v2/node_modules
```
LaRiffle: Change path?

client/.env
```diff
@@ -0,0 +1,4 @@
+REACT_APP_ENGINE_BACKEND_URL=https://engine.arkhn.org
+REACT_APP_HTTP_BACKEND_URL=http://localhost:4000
+REACT_APP_WS_BACKEND_URL=ws://localhost:4000
+REACT_APP_CLEANING_SCRIPTS_URI=http://localhost:5000
```
LaRiffle: Will be changed to 5001 (conflict port with fhir api)

client/src/components/Navbar/index.tsx
```diff
@@ -1,83 +1,43 @@
...
+import useReactRouter from 'use-react-router';
 
-import Drawer from "./components/Drawer";
-import Header from "./components/Header";
+import Drawer from './drawer';
+import Header from './header';
 
-import { changeNode } from "../../services/selectedNode/actions";
-import { login, logout } from "../../services/user/actions";
+import { logout } from '../../services/user/actions';
```
LaRiffle: Can we put absolute path?
`import { logout } from 'src/services/user/actions';`

client/src/components/authentication/index.tsx
```diff
@@ -1,35 +1,32 @@
...
+import { Mutation } from 'react-apollo';
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
+import { loader } from 'graphql.macro';
 
-import Navbar from "../../components/Navbar";
+import Navbar from '../navbar';
 
-import { login as loginAction } from "../../services/user/actions";
+import { login as loginAction } from '../../services/user/actions';
```
LaRiffle: same suggestion to use absolute path when possible

client/public/index.html
simonvadee: don't know, probably useless but it was there before

.github/workflows/server-v2.yml
simonvadee: done!

client/.env
simonvadee: thanks, done.

client/src/components/Navbar/index.tsx
simonvadee: done.

client/src/components/authentication/index.tsx
simonvadee: done.

PR #104, opened 2020-01-13
Update docker-compose and github workflows

LaRiffle:


client/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:10.15-alpine
```
simonvadee: you're testing on node 12 in the CI, you should probably use the same version here

client/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+version: "3"
+services:
+  pyrog_client:
+    image: arkhn/pyrog-client:latest
+    container_name: pyrog-client
+    build:
+      context: .
+    ports:
+      - $PORT:$PORT
+    env_file: .env
```
simonvadee: I don't understand how this works: the `.env` file points to `localhost` for the backend url, how does this work in docker ?

simonvadee: we should probably have both a `.env` and a `.env.docker` (same as in the server) ?

client/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+WORKDIR /usr/src/app
+COPY package.json yarn.lock ./
+RUN yarn install
+COPY . .
+EXPOSE 1729
+ENV NODE_ENV docker
+CMD ["yarn", "start"]
```
simonvadee: I think the proper way of doing it is to do `yarn build` to produce a "production" build and then do `node build/index.js`.
However while trying that I had issues with env variables (because yarn start uses the `.env` file by default, which `node` does not do) and it's also harder to debug because the errors reference the "compiled" js files rather than the actual typescript files. You can leave it like that for now I guess.

client/Dockerfile
LaRiffle: Yes indeed - any suggestion on which version we should use in general?

LaRiffle:
@simonvadee do you know why we have one pending test? 🧐
> Build Pyrog UI Expected — Waiting for status to be reported

client/docker-compose.yml
LaRiffle: :+1: 
I don't really know why this worked indeed^^

.github/workflows/client.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Build
         run: |
           cd client/
-          yarn build:docker
+          yarn build#:docker
```
simonvadee: typo

.github/workflows/client.yml
```diff
@@ -48,14 +48,14 @@ jobs:
       TAG: latest
     runs-on: ubuntu-18.04
     needs: build
-    if: github.ref == 'refs/heads/master'
+    #if: github.ref == 'refs/heads/master'
```
simonvadee: uncomment

.github/workflows/server.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       TAG: latest
     runs-on: ubuntu-18.04
     needs: build
-    if: github.ref == 'refs/heads/master'
+    #if: github.ref == 'refs/heads/master'
```
simonvadee: uncomment

client/.env.docker
```diff
@@ -0,0 +1,4 @@
+PORT=1729
+REACT_APP_HTTP_BACKEND_URL=http://pyrog-server:4000
+REACT_APP_WS_BACKEND_URL=ws://pyrog-server:4000
+REACT_APP_CLEANING_SCRIPTS_URI=http://cleaning-scripts:5001
```
simonvadee: can you add a cleaning-script service in the docker-compose.yml now that you setup it ?

server/package.json
```diff
@@ -1,5 +1,7 @@
 {
-  "name": "pyrog",
+  "name": "pyrog-version",
+  "version": "0.2.0",
+  "license": "GPL-3.0",
```
simonvadee: 👍

.github/workflows/client.yml
LaRiffle: Yeah I need putting only `yarn build` because that's all that's defined in package.json for the client

client/.env.docker
LaRiffle: Yeah I will!

.github/workflows/client.yml
```diff
@@ -30,3 +34,36 @@ jobs:
         run: |
           cd client
           yarn build
+
+#      - name: Lint
```
simonvadee: either remove these linesor add a comment to explain why it's commented out

client/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+
+version: "3"
+services:
+  pyrog_client:
+    image: arkhn/pyrog-client:latest
+    container_name: pyrog-client
+    build:
+      context: .
+    ports:
+      - $PORT:$PORT
```
simonvadee: does this use the PORT variable from the .env.docker file  ?

LaRiffle: Yes :)

simonvadee: :+1:

PR #115, opened 2020-01-17
Generate FHIR definitions from FHIR Bundle files

simonvadee:
This introduce a major change in Pyrog, it's the first PR related to FHIR profiles 😱 

The first step we're heading towards is moving away from the JSON-schema approach and use FHIR `StructureDefinition` instead 🎉 

In this PR I added the FHIR definitions published by HL7 (in `server/fhir/bundles`) and a script able to parse these bundles and turn them into StructureDefinition JSON files, which we will later use to represent the structure of all resources.


server/prisma/schema.prisma
```diff
@@ -45,8 +49,38 @@ model Credential {
...
+// enum FhirKind {
+//   logical
+//   primitive-type
+//   complex-type
+//   resource
+// }
+
+enum FhirDerivation {
+  constraint
+  specialization
```
Jasopaum: Isn't it better to have upper case enum values?

simonvadee: yes definitely, but then we need to change the actual `derivation` field from the StructureDefinition (which is lowercase). what do you think is best ?

server/scripts/fhirDefinitions.ts
```diff
@@ -0,0 +1,77 @@
...
+import { join } from 'path'
+
+import { Photon, StructureDefinitionCreateInput } from '@prisma/photon'
+
+import { HL7_AUTHOR } from '../src/constants'
+
+const photon = new Photon()
+
+const parseBundles = async (directory: string) => {
+  const dir = await fs.promises.opendir(directory)
```
Jasopaum: Maybe we should check at some point that we process only the json files we want to process.
Or maybe we can add a subfolder in server/fhir, just in case we need to have something else in this folder in the future

server/scripts/fhirDefinitions.ts
```diff
@@ -0,0 +1,77 @@
...
+      }),
+    )
+  }
+}
+
+const countDefinitions = async () => {
+  // resources
+  const resources = await photon.structureDefinitions.findMany({
+    where: { derivation: 'specialization', kind: 'resource' },
+    select: { id: true },
```
Jasopaum: Just out of curiosity: do we need to have the select field not empty?

server/prisma/schema.prisma
Jasopaum: Hum, right, or we could capitalize the string in between (something like `derivation: resource.derivation.toUpperCase()`) but I'm not sure if we should do that or not... We don't want this .toUpperCase() at several places

server/scripts/fhirDefinitions.ts
simonvadee: we don't "need" it, but this query just counts the number of records in the DB so I don't want to overfetch (if select is not empty, all fields are fetched)

Jasopaum: You mean "if select **is** empty, all fields are fetched"?
Thanks, this answers my question 👍

server/scripts/fhirDefinitions.ts
simonvadee: good remark, now you specify a list of files to parse instead of a folder

PR #116, opened 2020-01-17
Add module to cache FHIR structureDefinitions

Jasopaum:


server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
+const nodeCache = require('node-cache')
```
simonvadee: use the `import` syntax instead of `require`

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  // Put obj in key value store
+  cache.set(key, struct)
+}
+
+export const getStructureDefinition = (key: string) => cache.get(key)
+
+const buildCustomStructure = (structResource: any): any => {
+  if (!('snapshot' in structResource)) {
```
simonvadee: the right way of doing this is `if (!structResource.snapshot) ...`

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const cache = new nodeCache()
+
+export const setStructureDefinition = (fhirStructureDefinition: any) => {
+  const struct = buildCustomStructure(fhirStructureDefinition.resource)
+
+  // Use id as key. If it isn't present, use url
+  const key = struct.id || struct.url
+
+  if (key === undefined) {
```
simonvadee: if (!key)

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+    throw new Error('Structure definition has no id nor url field.')
+  }
+
+  // Put obj in key value store
+  cache.set(key, struct)
```
simonvadee: does this returns a promise ?

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const buildCustomStructure = (structResource: any): any => {
+  if (!('snapshot' in structResource)) {
+    throw new Error('Snapshot is needed in the structure definition.')
+  }
+
+  // Create the new custom structure
+  var customStruct = {}
+
+  Object.keys(structResource).forEach(key => {
```
simonvadee: `Object.keys(structResource).filter(structureFieldsWhiteList.includes).forEach(...)` to avoid the `if`

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+    throw new Error('Structure definition has no id nor url field.')
+  }
+
+  // Put obj in key value store
+  cache.set(key, struct)
+}
+
+export const getStructureDefinition = (key: string) => cache.get(key)
+
+const buildCustomStructure = (structResource: any): any => {
```
simonvadee: I would rename `buildCustomStructure` to `structuralize` ("that puts into structure") and `structResource` to `definition`. Does that make sense ?

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+  var customStruct = {}
+
+  Object.keys(structResource).forEach(key => {
+    if (structureFieldsWhiteList.indexOf(key) === -1) return
+    _.set(customStruct, key, structResource[key])
+  })
+
+  // If the structure defines a primitive type (one which we don't need to unroll in UI)
+  // we don't need the properties field
+  if (structResource.kind !== 'primitive-type') {
```
simonvadee: what happens if the definition is a `primitive-type` indeed ? it's not even in the final structure right now

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+    structResource.snapshot.element.forEach((element: any, index: number) => {
+      // Don't process first element in snapshot, it's the root
+      if (index == 0) return
```
simonvadee: by doing that you're missing the cardinality of the definition (it's defined in the first element)

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      const elementName = element.id.substring(element.id.indexOf('.') + 1)
+
+      // We skip some elements we don't need as id, extension
+      if (elementBlackList.indexOf(elementName) > -1) return
```
simonvadee: use `.contains` instead of `.indexOf`

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      // Create custom element
+      var customElement = {}
+
+      Object.keys(element).forEach(key => {
```
simonvadee: same comment as above: use Array.filter

server/src/fhir/cacheStructureDefinitions.ts
```diff
@@ -0,0 +1,72 @@
...
+const structureFieldsWhiteList = [
+  'id',
+  'url',
+  'name',
+  'description',
+  'kind',
+  'baseDefinition',
+  'derivation',
+]
+const elementFieldsWhiteList = ['definition', 'min', 'max']
```
simonvadee: you definitely need the `type` property here!  as well as the constraints maybe ?

server/src/fhir/cacheStructureDefinitions.ts
Jasopaum: Did you mean `.includes`?

server/src/fhir/cacheStructureDefinitions.ts
Jasopaum: No, it returns a boolean

server/src/fhir/cacheStructureDefinitions.ts
Jasopaum: That's right! But do you think we need it in Pyrog?

server/src/fhir/cacheStructureDefinitions.ts
Jasopaum: If it's a primitive type, we already have basic information in structure as name, description, ...
I don't think we need something more in Pyrog, do we?

server/src/fhir/cacheStructureDefinitions.ts
Jasopaum: I added `type` and `constraints`

server/src/fhir/cacheStructureDefinitions.ts
simonvadee: yes sorry...

server/src/fhir/cacheStructureDefinitions.ts
simonvadee: :+1:

server/src/fhir/cacheStructureDefinitions.ts
simonvadee: well the cardinality is important in order to know whether an element is a list (0..*), whether it's required (1..1) or not (0..0) etc... so yes I think we do need it.

PR #136, opened 2020-02-05
Production docker image for pyrog-client

simonvadee:


client/Dockerfile
```diff
@@ -6,8 +6,10 @@ RUN yarn install
 
 COPY src src
 COPY public public
+COPY .env.staging .env.staging
 
-EXPOSE 1729
-ENV NODE_ENV docker
+ENV NODE_ENV production
```
Jasopaum: production or staging?

server/src/server.ts
```diff
@@ -29,11 +29,14 @@ const options = {
...
 }
 
+const { PORT } = process.env
+
 const main = async () => {
   await bootstrapDefinitions()
   await server.start(options, () =>
     console.log(
-      `🚀 Server ready at: http://localhost:4000\n⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️⭐️`,
+      `🚀 Server ready at: http://localhost:${PORT || 4000}
```
Jasopaum: In which case PORT would be undefined?
Didn't you change the server port from 4000 to 1000?

PR #142, opened 2020-02-07
Move StructureDefinitions to fhir-api

simonvadee:


server/src/fhir/definitions.ts
```diff
@@ -28,24 +27,38 @@ export const bootstrapDefinitions = async () => {
     }
 
     // Put obj in key value store
+    console.log('loaded', id)
```
Jasopaum: You could remove this

server/src/resolvers/mapping/index.ts
```diff
@@ -4,9 +4,11 @@ import {
   MAPPING_VERSION_1,
   MAPPING_VERSION_2,
   CURRENT_MAPPING_VERSION,
+  MAPPING_VERSION_3,
 } from '../../constants'
 import handleV1 from './v1'
 import handleV2 from './v2'
+import handleV3 from './v2'
```
Jasopaum: from './v2'

client/src/components/mapping/FhirMappingPanel/AddResource/index.tsx
```diff
@@ -96,17 +90,30 @@ const AddResource = ({ callback }: Props) => {
     }
   );
 
+  React.useEffect(() => {
+    async function fetchDefinitions() {
```
Jasopaum: T'aimes pas ce que t'as fait

server/src/fhir/definitions.ts
```diff
@@ -28,24 +27,38 @@ export const bootstrapDefinitions = async () => {
...
+      params: {
+        _count: 1000, // TODO: use pagination
+      },
+    },
+  )
+  for (const def of standardDefinitions.items.map(
+    ({ _source }: any) => _source,
+  )) {
+    cacheDefinition(def)
   }
```
Jasopaum: Would something like 
```
standardDefinitions.items
  .map(({ _source }: any) => _source)
  .map(d => cacheDefinition(d))
```
be clearer?

simonvadee: yes!

server/src/resolvers/mapping/index.ts
simonvadee: done

server/src/fhir/definitions.ts
simonvadee: i prefer to keep this for now

PR #148, opened 2020-02-11
Select resource profile

simonvadee:
We can now select a profile for a resource using a (filterable) sub-menu

### DIFF
--- a/client/src/components/selects/TSelect.tsx
+++ b/client/src/components/selects/TSelect.tsx
@@ -11,6 +11,8 @@ import * as React from 'react';
 interface ISelectProps<T> {
   disabled: boolean;
   displayItem: (item: any) => string;
+  className?: string;
+  rightIcon?: IconName;
   sortItems?: ItemListPredicate<T>;
   filterItems: ItemPredicate<T>;
   filterable?: boolean;
@@ -35,6 +37,8 @@ export default class TSelect<T> extends React.Component<ISelectProps<T>, any> {
     const {
       disabled,
       displayItem,
+      className,
+      rightIcon,
       sortItems,
       filterItems,
       filterable,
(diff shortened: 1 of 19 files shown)


client/src/components/selects/TSelect.tsx
```diff
@@ -61,10 +66,12 @@ export default class TSelect<T> extends React.Component<ISelectProps<T>, any> {
       >
         <Button
           disabled={disabled}
+          fill={true}
+          alignText={'left'}
```
Jasopaum: Does it change the display somewhere else?
Maybe we could also pass these as optional props to TSelect

client/src/components/selects/addResourceSelect.tsx
```diff
@@ -5,96 +5,125 @@ import { IconName } from '@blueprintjs/icons';
...
+            <strong>{profile.name}</strong>
           </span>
         }
       />
     );
   };
 
-  private filterByName: ItemPredicate<string> = (query, resource: string) => {
-    return resource.toLowerCase().indexOf(query.toLowerCase()) >= 0;
+  const renderItem: ItemRenderer<Resource> = (
```
Jasopaum: It's a detail but I think it could be nice to have something to distinguish the base resource from the others, or at least to be sure that it's at the top of the list

client/src/components/selects/TSelect.tsx
simonvadee: It does change the display elswhere but it's actualy nicer so I'll keep it that way until we need to make this configirable, how does that sound ?

client/src/components/selects/addResourceSelect.tsx
simonvadee: good point, I'll add that

client/src/components/selects/TSelect.tsx
Jasopaum: If it makes things nicer, I'm good with that 👍

client/src/components/selects/addResourceSelect.tsx
simonvadee: done

PR #157, opened 2020-02-12
Cache definitions in redis

simonvadee:


server/src/fhir/definitions.ts
```diff
@@ -1,38 +1,64 @@
...
 
 // Gets a definition from the cache.
-export const getDefinition = (key: string): StructureDefinition | undefined =>
-  cache.get(key)
+export const getDefinition = async (
+  key: string,
+): Promise<StructureDefinition | undefined> => {
+  const { get } = cache()
+  const res = await get(key)
+  return res ? JSON.parse(res) : undefined
```
Jasopaum: Are you doing this because we can't put objects in the cache?

server/src/resolvers/Query.ts
```diff
@@ -101,7 +102,18 @@ export const Query = queryType({
         definitionId: idArg({ nullable: false }),
       },
       nullable: true,
-      resolve: (_, { definitionId }) => ({ id: definitionId }),
+      resolve: async (_, { definitionId }) => {
+        console.error('before')
```
Jasopaum: remove this

server/src/resolvers/Query.ts
```diff
@@ -101,7 +102,18 @@ export const Query = queryType({
...
+        const def = await getDefinition(definitionId)
+        console.error(!!def)
```
Jasopaum: and this

server/src/fhir/definitions.ts
simonvadee: yes!

server/src/resolvers/Query.ts
simonvadee: done

server/src/resolvers/Query.ts
simonvadee: done

PR #172, opened 2020-02-18
Add more info about attribute in redux store

Jasopaum:
This permitted to display concept map selection only when the attribute is a code


client/src/components/mapping/InputColumn/index.tsx
```diff
@@ -215,31 +213,33 @@ const InputColumn = ({ input, schema, source }: Props) => {
...
-                        variables: {
-                          inputId: input.id,
-                          data: { conceptMapId: null }
-                        }
-                      });
-                    }}
-                  />
-                </ButtonGroup>
-              </div>
+              {attribute.types[0] === 'code' ? (
```
simonvadee: I think there you can use the `&&` syntax : `{attribute.types[0] === 'code' && (...)}`

client/src/types.tsx
```diff
@@ -46,7 +48,7 @@ export interface ISelectedResource {
   };
 }
 
-export interface ISelectedAttribute {
+export interface ISelectedAttribute extends Node {
```
simonvadee: drop the I

PR #181, opened 2020-02-21
Add button to upload code systems

Jasopaum:


client/src/components/mapping/ConceptMap/index.tsx
```diff
@@ -9,6 +15,10 @@ import CodeSystemSelect from 'components/selects/codeSystemSelect';
 import StringSelect from 'components/selects/stringSelect';
 import { IReduxStore } from 'types';
 
+import UploadCodeSystem from 'components/uploads/uploadCodeSystem';
+import { validator } from 'components/uploads/validate';
```
simonvadee: maybe we should put that in the `service` folder instead of `components` now that we use it in multiple components ?

client/src/components/mapping/ConceptMap/index.tsx
```diff
@@ -474,81 +484,86 @@ const ConceptMapDialog = ({
...
+            <div className="center-text">
+              <h1>CONCEPT MAP</h1>
+            </div>
+            {!!selectedTargetSystem && !!sourceSystemTitle && metaData()}
+            <table className="bp3-html-table">
+              <thead>
+                <tr>
+                  <th className="head-col"></th>
+                  <th className="source-col">
+                    Source{':  '}
```
simonvadee: I'm not gonna say anything

client/src/components/uploads/uploadProfile/index.tsx
```diff
@@ -55,6 +56,8 @@ const UploadProfile = ({ isOpen, resource, onClose, onUpload }: Props) => {
   );
   const [uploadingProfile, setUploadingProfile] = React.useState(false);
 
+  const validate = validator(fhirSchema);
```
simonvadee: this will re-compile the json schema each time the component is re-rendered, I would prefer having it outside of the component.

client/src/components/uploads/uploadCodeSystem/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { validator } from 'components/uploads/validate';
+import fhirSchema from 'components/uploads/uploadCodeSystem/CodeSystem.schema.json';
+
+import './style.scss';
+
+const UploadCodeSystem = (): React.ReactElement => {
+  const toaster = useSelector((state: IReduxStore) => state.toaster);
+
+  const [newCodeSystem, setNewCodeSystem] = useState(undefined as any);
+  const [uploading, setUploading] = useState(undefined as any);
```
simonvadee: this is a boolean

client/src/components/uploads/uploadCodeSystem/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [invalidFile, setInvalidFile] = useState(false);
+
+  const validate = validator(fhirSchema);
+
+  const validateCodeSystem = (codeSystem: any): boolean => {
```
simonvadee: I don't think you need this function.. can't you use `validate` directly ? it returns a boolean

Jasopaum: That's right! I tried to keep the same structure as in uploadProfile but I don't do anything anymore in this function. I'll remove it

client/src/components/mapping/ConceptMap/index.tsx
Jasopaum: Currently, this folder is used for redux reducers and actions (maybe we can find another name)?
Do you want to get only the `components/uploads/validate.ts` file out of `components`?

simonvadee: you're right it would not fit into the services folder. leave it here sinc this is used only in uploads...

PR #184, opened 2020-02-26
Fix: create "schemas" directory on first use

achauve:


achauve:
The first commit is an unrelated refactoring, as in local I had not run prisma migrate so the graphql query was failing and so the front. So I tried a fix before understanding it was a migration problem. 

Do you want me to delete this commit? still it's a usefull nano refactoring I think.

simonvadee:
100% agree with the small refacto, let's go!

Issue #175, opened 2020-02-19
Create a backup for each source every time we do a migration


simonvadee:
there is a `yarn migrate` script that applies new prisma migrations to the postgres database. The prisma migration engine (called "lift") is still pretty "alpha" and it often causes data loss (been there).

The idea here is to export the mapping of all sources as JSON files and to save them in a docker volume before a migration is applied. This will allow us to re-import these backups in case some data is lost during the migration.

There is already a script to import a mapping (see server/scripts/importMapping.ts) but not to export one. However, there is already a resolver to export a mapping (server/src/resolvers/mapping/index.ts) that can be used to write such a script.

PR #187, opened 2020-02-28
[server] feat: Create sources backup before each migration

achauve:
#175


server/package.json
```diff
@@ -17,6 +17,7 @@
     "seed:superuser": "env $(cat .env) ts-node scripts/superuser.ts",
     "seed:mapping": "env $(cat .env) ts-node scripts/importMapping.ts",
     "seed:definitions": "env $(cat .env) ts-node scripts/fhirDefinitions.ts $(ls fhir/*.json)",
+    "export:mapping": "env $(cat .env) ts-node scripts/exportMapping.ts",
```
simonvadee: I would call it "export:mappings" (plural) since we consider that each source has its own mapping.
Also, I would add this command before the "yarn migrate"

server/scripts/exportMapping.ts
```diff
@@ -0,0 +1,33 @@
...
+import { exportMapping } from '../src/resolvers/mapping'
+
+const photon = new Photon()
+
+const main = async (outputDirectory: string) => {
+  if (!existsSync(outputDirectory)) {
+    mkdirSync(outputDirectory)
+  }
+
+  let sources = await photon.sources.findMany()
```
simonvadee: I think this ca be a `const`

server/scripts/exportMapping.ts
```diff
@@ -0,0 +1,33 @@
...
+  for (const source of sources) {
+    console.log('source:', source)
+    const content = await exportMapping(photon, source.id)
+    writeFileSync(`${outputDirectory}/${source.id}.json`, content)
```
simonvadee: I would use ${source.template.name}_${source.name}.json as file name in order for it to be recognizable.

simonvadee: (there can't be two sources with the same name for a single template)

server/package.json
achauve: yes

server/scripts/exportMapping.ts
achauve: yes

server/scripts/exportMapping.ts
achauve: Why not, but does it need to be recognizable? as a general rule I prefer not to use dynamic entity names as filenames as they can contain any character.

simonvadee: maybe not because we'll probably want to re-import all sources if we lose data but it might come in handy. Leave it like this for now, we'll see if we ever need it to be human-readable!

PR #190, opened 2020-02-28
Augment mapping export with definition

simonvadee:
fhir-pipe needs the definition and source id in the resource mapping

### DIFF
--- a/server/src/resolvers/mapping/index.ts
+++ b/server/src/resolvers/mapping/index.ts
@@ -56,6 +56,7 @@ export const exportMapping = async (
   let resources = await photon.resources({
     where: { source: { id: source.id } },
     include: {
+      source: true,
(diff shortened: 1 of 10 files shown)


server/src/resolvers/mapping/index.ts
```diff
@@ -64,6 +69,23 @@ export const exportMapping = async (
...
+  // augment the mapping of each resource with its definition
+  // and its source id. This is now required by fhir-pipe when
+  // running from a mapping file.
+  resources = await Promise.all(
+    resources.map(async r => {
+      const def = await getDefinition(r.definitionId)
+      return {
+        ...r,
+        definition: def?.$meta,
+        source: {
```
Jasopaum: Can't the source id be included in `photon.resources()`?

PR #193, opened 2020-03-02
Add way to filter rows to map

### LINKED ISSUE #173: Handle logic on mapping pre-processing
Sometimes, a logical filtering of data needs to be done before mapping. 

- MIMIC has a table chartevents where all the observations are put, whatever their type. Chartevents have a code `item_id` which identifies the observation type: blood pressure, weight etc.. We want to be able to filter chartevents with this code, before we map only the blood pressure items inside the Observation-bloodpressure resource. 

- On Pyrog, we could add it in the same menu where we pick the primary key of the resource, then add a filterColumn button to select the column to be filtered. 
- This requires handling filtering with fhir-pipe, by enhancing the SQL query like `select * from observationTable where primaryKey=key and filterColumn=filterCode`

Jasopaum:
Closes #173 

The UI looks like (in the resource selection drawer):

Capture d’écran 2020-03-02 à 18 39 23


client/src/components/mapping/FhirMappingPanel/ResourceSelector/Drawer/index.tsx
```diff
@@ -181,17 +191,159 @@ const Drawer = ({
     }
   );
 
+  const pickPrimaryKay = (
```
simonvadee: typo: Kay

client/src/components/mapping/FhirMappingPanel/ResourceSelector/Drawer/index.tsx
```diff
@@ -181,17 +191,159 @@ const Drawer = ({
...
+  );
+
   return (
     <BPDrawer
       title={resource.definition ? resource.definition.type : ''}
       icon={resource.definition ? 'properties' : null}
       isOpen={isOpen}
       onClose={onCloseCallback}
-      size={BPDrawer.SIZE_SMALL}
+      // size={BPDrawer.SIZE_SMALL}
```
simonvadee: comment

server/src/generated/schema.graphql
```diff
@@ -270,6 +302,7 @@ type Resource {
   """Structured version of a definition"""
   definition: StructureDefinition!
   definitionId: String!
+  filters(after: ID, before: ID, first: Int, last: Int, skip: Int): [Filter!]!
```
simonvadee: filters should not be searchable

server/src/resolvers/mapping/utils.ts
```diff
@@ -70,3 +73,16 @@ export const buildAttributesQuery = (
...
+
+const buildColumnWithoutJoinsQuery = (
+  c: Column,
+): ColumnCreateWithoutInputInput => clean(c)
+
+export const buildFiltersQuery = (
+  filters: FilterWithSqlColumn[],
+): FilterCreateInput[] | null =>
+  filters.map(f => {
+    let filter: FilterCreateInput = clean(f)
```
simonvadee: can you use const everywhere ?

PR #205, opened 2020-03-23
Add read/write permissions

### LINKED ISSUE #199: Create a validation mode on pyrog
Create a validation mode on pyrog, where the user can browse through the mapping and add comments, but cannot add or delete attributes.

Jasopaum:
Closes #199


server/src/resolvers/Resource.ts
```diff
@@ -162,3 +162,23 @@ export const updateResource: FieldResolver<
     data,
   })
 }
+
+export const resourcesForUser: FieldResolver<
```
simonvadee: I think we can get rid of this as discussed

simonvadee: We still need to filter the resources though, so you can add the filter on the source.resources resolver

server/src/resolvers/Query.ts
```diff
@@ -74,16 +85,16 @@ export const Query = queryType({
         }),
     })
 
-    t.list.field('resources', {
+    t.list.field('resourcesForUser', {
```
simonvadee: let's get rid of this

server/src/permissions/index.ts
```diff
@@ -15,13 +101,53 @@ const rules = {
     })
     return Boolean(user && user.role == 'ADMIN')
   }),
+  isWriter: rule()(async (_, args, ctx: Context) => {
```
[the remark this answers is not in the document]
simonvadee: also, return true if the user is a global admin

client/src/graphql/queries/resourcesForUser.graphql
```diff
@@ -0,0 +1,10 @@
+query resourcesForUser($userId: ID!, $filter: ResourceWhereInput) {
```
simonvadee: drop this once you removed it from the back

server/src/resolvers/Source.ts
```diff
@@ -53,7 +53,7 @@ export const Source = objectType({
 
 export const createSource: FieldResolver<'Mutation', 'createSource'> = async (
   _parent,
-  { templateName, name, hasOwner, mapping },
+  { templateName, name, hasOwner, userId, mapping },
```
simonvadee: you get the userId from the context, you don't need to pass it to the mutation

client/src/graphql/mutations/createSource.graphql
```diff
@@ -2,12 +2,14 @@ mutation createSource(
   $templateName: String!
   $name: String!
   $hasOwner: Boolean!
+  $userId: ID!
```
simonvadee: no need, see below

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -86,51 +88,68 @@ const StaticValueForm = ({ attribute }: Props) => {
...
+              resourceId: resource.id,
+              definitionId: curNode.types[0],
+              path: parentPath
+            }
+          });
+          dispatch(setAttributeInMap(parentPath, attr.createAttribute));
+        }
       }
-    });
+      createStaticInput({
```
simonvadee: why don't we await this ?

client/src/components/mapping/TabColumnPicking/DynamicColumnPicker/index.tsx
```diff
@@ -61,10 +62,24 @@ const DynamicColumnPicker = ({ attribute, schema, source }: Props) => {
     undefined as number | undefined
   );
 
-  const [createAttribute] = useMutation(mCreateAttribute);
-  const [createSQLInput, { loading: creatingSQLInput }] = useMutation(
-    mCreateSQLInput
-  );
+  const onError = (error: ApolloError): void => {
```
simonvadee: can't this be factorized ?

client/src/components/mapping/FhirMappingPanel/AddResource/index.tsx
```diff
@@ -106,18 +103,19 @@ const AddResource = ({ callback }: Props) => {
     });
 
     // Create the new resource
-    const {
-      data: { createResource: createdResource }
-    } = await createResource({
+    const responseCreateResource = await createResource({
```
simonvadee: you can destructure `data` here, it will still exist even if an error is returned i think

client/src/components/fhirpipe/index.tsx
```diff
@@ -29,14 +29,15 @@ interface Resource {
 }
 
 // GRAPHQL
-const qSources = loader('src/graphql/queries/sources.graphql');
-const qResources = loader('src/graphql/queries/resources.graphql');
+const qSources = loader('src/graphql/queries/sourcesForUser.graphql');
+const qResources = loader('src/graphql/queries/resourcesForUser.graphql');
```
simonvadee: drop this (see below)

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
Jasopaum: Do we need to if nothing is done after?

client/src/components/mapping/TabColumnPicking/DynamicColumnPicker/index.tsx
Jasopaum: I'd love to! I'm just not sure where to put it...

client/src/components/mapping/FhirMappingPanel/AddResource/index.tsx
Jasopaum: I tried it and I think the whole mutation returns undefined if an error is caught in the middle (and the onError function is called)

server/prisma/schema.prisma
```diff
@@ -144,4 +144,16 @@ model User {
 
   updatedAt DateTime @updatedAt
   createdAt DateTime @default(now())
+}
+
+enum Rights {
```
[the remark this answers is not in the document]
Jasopaum: I changed Rights to SourceRole.
But I'm not sure if READER and WRITER should go along with USER and ADMIN in the same enum

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
simonvadee: if error handling is not affected, it's fine

server/src/permissions/index.ts
```diff
@@ -101,13 +23,20 @@ const rules = {
     })
     return Boolean(user && user.role == 'ADMIN')
   }),
-  isWriter: rule()(async (_, args, ctx: Context) => {
+  isSourceWriter: rule()(async (_, args, ctx: Context) => {
     // Get user id
     const userId = getUserId(ctx)
 
+    // Return true if the user is admin
+    const user = await ctx.photon.users.findOne({
```
simonvadee: just a thought for later: multiplying the db calls in an acl middleware impacts dthe performance badly, now that we have redis set up, we should probably cache the logged in users in there so we have a fast way to access user details

server/src/permissions/index.ts
```diff
@@ -156,24 +89,29 @@ export const permissions = shield({
...
-    deleteResource: rules.isWriter,
+    createResource: rules.isSourceWriter,
+    updateResource: rules.isSourceWriter,
+    deleteResource: rules.isSourceWriter,
 
-    createAttribute: rules.isWriter,
-    updateAttribute: rules.isWriter,
+    createAttribute: rules.isSourceWriter,
+    updateAttribute: rules.isSourceWriter,
     updateComments: rules.isAuthenticatedUser,
```
simonvadee: shouldn't this be `isSourceReader` ?

server/src/permissions/index.ts
Jasopaum: I'll write an issue for that!

server/src/permissions/index.ts
Jasopaum: In my mind, everybody authenticated could update comments. Do you think I should add a new rule?

PR #210, opened 2020-03-25
 Integrate fhir.ts in pyrog client and server 

simonvadee:


client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -84,137 +67,122 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
-    // Then we find the parent from which we want to remove
-    // the child node.
-    const parent = findNode(stateNodes, path.parent!);
-    if (!parent || !parent.childNodes) return stateNodes;
+    dispatch(removeAttributesFromMap(deleted.path));
+    // and remove the item in the parent attribute as well
+    array.removeItem(deleted);
 
-    if (parent.childNodes.length === 1) {
+    if (arrayNode.childNodes?.length === 1) {
```
Jasopaum: Shouldn't you check that the length of the array is 0 now that you removed the element?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -84,137 +67,122 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
-      parent.childNodes[parent.childNodes.length - 1].nodeData?.index! + 1;
-
-    // build a new child node
-    const childNode = new Node(path, path.definition, nextIndex);
-    const newNode = createNode(
-      childNode,
-      genTreeLevel(contentChildren || [], childNode),
-      undefined,
-      () => setNodes(nodes => deleteNodeFromArray(nodes, childNode))
+  const newItemNode = (arrayNode: ITreeNode<Attribute>, index?: number) => {
```
Jasopaum: Do you think the Attribute class could have a method addChild, createChild or something like that?
It could be called on a Attribute which is an array and return a new child or add it directly to it children.

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -225,121 +193,77 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
+    });
+    if (!data || !data.structureDefinition) return [];
+    return data.structureDefinition.attributes.map(Attribute.from);
   };
 
-  const handleNodeClick = async (node: ITreeNode<Node>): Promise<void> => {
-    if (node.nodeData?.isArray) {
+  const handleNodeClick = async (node: ITreeNode<Attribute>): Promise<void> => {
+    const attribute = node.nodeData!;
+    console.debug(attribute.path);
```
Jasopaum: Don't forget this

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -225,121 +193,77 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
+    if (!attribute.isPrimitive) {
+      // if the node is of composite type, expand (or collapse) it
       node.isExpanded = !node.isExpanded;
-    } else if (!node.nodeData?.isPrimitive) {
+      // if the node has no children yet, fetch the attributes definitions
+      // and add them as children of the clicked node.
       if (!node.childNodes || node.childNodes.length === 0) {
         // TODO what if several types are possible?
```
Jasopaum: Can we remove this comment now?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
simonvadee: I removed the attribute (ie: arrayNode.nodeData) but not the childNode

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/nodeLabel.tsx
```diff
@@ -1,31 +1,30 @@
...
+    const menu = isArray ? (
       <Menu>
         <MenuItem
           icon={'add'}
           onClick={addNodeCallback}
           text={'Ajouter un item'}
         />
       </Menu>
-    ) : deleteNodeCallback ? (
+    ) : isItem && !isSlice ? (
```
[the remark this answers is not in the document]
simonvadee: sure. By definition, an array can have 1..N _items_ (not _children_). If the array have slices, slices are "propagated" (copied) to each children so that each item of the array can "implement" a slice.

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
simonvadee: the Attribute class already has such a method, but the newItemNode function is about adding an ITreeNode<Attribute> so that wouldn't fit into the Attribute class. Did I get your point ?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
simonvadee: does it bother you if I keep it for now ? I find it convenient to understand what's going on (even on staging it might be useful)

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
simonvadee: yes! done

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
Jasopaum: I'm not sure. I just wanted to avoid the `array.definition.max = '1'/'*';` hack to be done here

simonvadee: I think that's better indeed, i'll try that!

PR #212, opened 2020-03-25
Hot fix codesystems

Jasopaum:
Prevent Pyrog from crashing when some CodeSystems don't have a title.
Ideally, we'd like a better solution but we need to have Pyrog functional to prepare a demo.

### DIFF
--- a/client/src/components/selects/codeSystemSelect.tsx
+++ b/client/src/components/selects/codeSystemSelect.tsx
@@ -30,7 +30,8 @@ interface Props {
 }
 
 const filterByName: ItemPredicate<CodeSystem> = (query, item) => {
-  return `${item.title.toLowerCase()}`.indexOf(query.toLowerCase()) >= 0;
+  const title = item.title || item.name;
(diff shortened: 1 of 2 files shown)


client/src/components/selects/codeSystemSelect.tsx
```diff
@@ -30,7 +30,8 @@ interface Props {
 }
 
 const filterByName: ItemPredicate<CodeSystem> = (query, item) => {
-  return `${item.title.toLowerCase()}`.indexOf(query.toLowerCase()) >= 0;
+  const title = item.title || item.name;
```
elsiehoffet-94: Are we sure that there will always be either a title or a name? They aren't required attributes

Jasopaum: At least it works with the code systems we have now but it's indeed not a very robust solution...
I'll try to think about a better way to handle this

simonvadee: add //TODO comment

PR #217, opened 2020-03-27
Upgrade prisma to preview24

simonvadee:
I also updated the nexus dependencies to workj with the new version of prisma.
I encourage you read the release notes since the last version (which was preview19) --> https://github.com/prisma/prisma2/releases

There are some improvements to the prisma engine worth mentioning here:

```
Another improvement is a better debugging experience. When setting the DEBUG environment variable (e.g. with export DEBUG="*"), the logging output now contains the names of Prisma Client API calls.
```

```
you can now have tables:

    without primary keys (but unique constraints)
    with composite primary keys
    with composite unique constraints
```

```
Send raw SQL queries using prisma.raw(sql) (read the API docs)
Stopped shipping TypeScript with Prisma Client JS
```


server/prisma/schema.prisma
```diff
@@ -11,7 +11,7 @@ datasource db {
 model Template {
   id      String   @default(cuid()) @id
   name    String   @unique
-  sources Source[] @relation(onDelete: CASCADE)
```
Jasopaum: Didn't they add this feature yet? 😞

simonvadee: still no :'(

PR #219, opened 2020-03-27
Add unique constraint on (user, source) in ACL

Jasopaum:


Jasopaum:
> don't you want to add a unique constraint on source like: @@unique([template, name]) as well ? we're doing that with code right now

Done

PR #221, opened 2020-03-27
Admin can see all sources (and cache user in redis)

simonvadee:
user is cached in redis on login/signup
user is put in context for every request (if auth token is ok)

update the `sources` resolver so that admins can see all sources

### DIFF
--- a/server/src/cache.ts
+++ b/server/src/cache.ts
@@ -5,6 +5,7 @@ import { REDIS_URL } from './constants'
 
 let rcache: RedisClient
 let get: (arg: string) => Promise<string>
+let del: (arg: string) => Promise<number>
 let mget: OverloadedKeyCommand<string[], string[], Promise<string[]>>
 let set: (arg1: string, arg2: string) => Promise<unknown>
 let sadd: OverloadedKeyCommand<string, number, Promise<number>>
@@ -17,6 +18,7 @@ export default () => {
   if (rcache)
     return {
       get,
+      del,
       mget,
       set,
       sadd,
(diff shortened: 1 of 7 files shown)


server/src/resolvers/Source.ts
```diff
@@ -57,12 +57,13 @@ export const sources: FieldResolver<'Query', 'sources'> = async (
...
-  const userId = getUserId(ctx)
-  const accesses = await ctx.prisma.accessControl.findMany({
-    where: { user: { id: userId } },
-    include: { source: true },
+  const { role, id } = ctx.user!
+
+  if (role === 'ADMIN') return ctx.prisma.source.findMany()
+
+  return ctx.prisma.source.findMany({
+    where: { accessControls: { some: { user: { id } } } },
```
Jasopaum: Cool, didn't know we could do that 👍

server/src/cache.ts
```diff
@@ -5,6 +5,7 @@ import { REDIS_URL } from './constants'
 
 let rcache: RedisClient
 let get: (arg: string) => Promise<string>
+let del: (arg: string) => Promise<number>
```
Jasopaum: You finally didn't use this, did you?

simonvadee: no I didn't but now it's there when we need it :)

Jasopaum: What about YAGNI? ;)

simonvadee: what do you think the french government thinks about YAGNI now that we're out of masks and respirators ?

simonvadee: covid-19 should be an inspiration to all developers around the world

Jasopaum: ❤️

PR #223, opened 2020-03-30
Minor interface improvements

Jasopaum:
- Make tab div scrollable
- Concept map description can stay empty
- Prevent spaces in concept map title

### DIFF
--- a/client/src/components/mapping/ConceptMap/index.tsx
+++ b/client/src/components/mapping/ConceptMap/index.tsx
@@ -348,7 +348,9 @@ const ConceptMapDialog = ({
             className="right"
             onChange={(e: React.FormEvent<HTMLElement>): void => {
               const target = e.target as HTMLInputElement;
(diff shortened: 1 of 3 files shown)


client/src/components/mapping/ConceptMap/index.tsx
```diff
@@ -348,7 +348,9 @@ const ConceptMapDialog = ({
             className="right"
             onChange={(e: React.FormEvent<HTMLElement>): void => {
               const target = e.target as HTMLInputElement;
-              setConceptMapTitle(target.value);
+              if (/^\S*$/.test(target.value)) {
```
simonvadee: do not set the concept map title if it matches the regex ?

Jasopaum: It's on each keystroke, so it just doesn't doesn't add a space if the user tries to add one

PR #226, opened 2020-03-30
Allow fhirpipe route only to admin users

Jasopaum:


client/src/components/navbar/index.tsx
```diff
@@ -42,15 +51,17 @@ const Navbar = ({ exportMapping }: Props) => {
...
-        <Button
-          icon="flame"
-          minimal={true}
-          onClick={() => {
-            history.push('/fhirpipe');
-          }}
-        >
-          Run fhir-pipe
-        </Button>
+        {isAdmin ? (
```
simonvadee: use {isAdmin && \<Button\>} instead of a ternary

PR #229, opened 2020-03-31
Enable custom identifier system

Jasopaum:
It's now possible to add a custom identifier system in the mapping.
This custom system will look like "http://terminology.arkhn.org/sourceId/resourceId".
The mapper now can select source and resource that will be used as a referential to disambiguate references whose value do not belong to a standard system.

Capture d’écran 2020-03-31 à 11 35 12

Capture d’écran 2020-03-31 à 11 35 20

### DIFF
--- a/client/src/components/fhirpipe/index.tsx
+++ b/client/src/components/fhirpipe/index.tsx
@@ -1,6 +1,5 @@
 import axios from 'axios';
 import { Button, MenuItem, Switch } from '@blueprintjs/core';
-import { ItemPredicate, MultiSelect, Select } from '@blueprintjs/select';
 import React, { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { useQuery } from '@apollo/react-hooks';
@@ -9,6 +8,9 @@ import { loader } from 'graphql.macro';
 
 import Navbar from 'components/navbar';
 import { IReduxStore } from 'types';
+
+import SourceSelect from 'components/selects/sourceSelect';
+import ResourceMultiSelect from 'components/selects/resourceMultiSelect';
 
 import { FHIRPIPE_URL } from '../../constants';
 
@@ -36,7 +38,7 @@ const qSourcesAndResources = loader(
   'src/graphql/queries/sourcesAndResources.graphql'
 );
 
-const FhirpipeView = () => {
+const FhirpipeView = (): React.ReactElement => {
   const toaster = useSelector((state: IReduxStore) => state.toaster);
 
   const [selectedSource, setSelectedSource] = useState({} as Source);
@@ -54,9 +56,7 @@ const FhirpipeView = () => {
   const sources = data ? data.sources : [];
   const credentials = selectedSource.id ? selectedSource.credential : undefined;
   const credentialsMissing = !!selectedSource.id && !credentials;
-
-  const SourceSelect = Select.ofType<Source>();
-  const ResourceMultiSelect = MultiSelect.ofType<Resource>();
+  // const ResourceMultiSelect = MultiSelect.ofType<Resource>();
 
   const isItemSelected = (list: any[], item: any): boolean => {
     return list.includes(item);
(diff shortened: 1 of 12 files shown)


client/src/components/fhirpipe/index.tsx
```diff
@@ -54,9 +56,7 @@ const FhirpipeView = () => {
   const sources = data ? data.sources : [];
   const credentials = selectedSource.id ? selectedSource.credential : undefined;
   const credentialsMissing = !!selectedSource.id && !credentials;
-
-  const SourceSelect = Select.ofType<Source>();
-  const ResourceMultiSelect = MultiSelect.ofType<Resource>();
+  // const ResourceMultiSelect = MultiSelect.ofType<Resource>();
```
simonvadee: remove

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -34,24 +44,47 @@ interface Props {
...
-  const { resource } = useSelector((state: IReduxStore) => state.selectedNode);
+  const { source, resource } = useSelector(
+    (state: IReduxStore) => state.selectedNode
+  );
   const attributesForResource = useSelector(
     (state: IReduxStore) => state.resourceInputs.attributesMap
   );
 
+  // If the current attribute is the identifier of the resource and not a Reference identifier
+  const isIdentifierSystem = /^identifier\[\d+\]\.system$/.test(attribute.path);
```
simonvadee: you should put that part in the Attribute class from fhir.ts and not use a regex here

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -138,14 +178,96 @@ const StaticValueForm = ({ attribute }: Props) => {
       await createStaticInput({
         variables: {
           attributeId,
-          staticValue
+          staticValue: customSystem
```
simonvadee: It bothers me that we have to put this logic directly inside the component but I can't think of anywhere else to put it

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -138,14 +178,96 @@ const StaticValueForm = ({ attribute }: Props) => {
...
+
+  const handleSourceSelect = (source: Source): void => {
+    setSelectedSource(source);
+    setSelectedResource(undefined);
+  };
+  const handleResourceSelect = (resource: Resource): void => {
+    setSelectedResource(resource);
+  };
+
+  const renderCustomSystemInput = (): React.ReactElement => (
```
simonvadee: I think that should be an other component (alongside with selectedSource and selectedResource)

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -155,7 +277,7 @@ const StaticValueForm = ({ attribute }: Props) => {
       </div>
       <FormGroup labelFor="text-input" inline={true}>
         <ControlGroup>
-          {attribute.isReferenceType ? (
+          {attribute.definition.id === 'Reference.type' ? (
```
simonvadee: too many ternaries, can you find an other way to render it ?

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
Jasopaum: In the `selects` folder?
It's not really a select but I agree that this could be somewhere else

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
Jasopaum: Agreed, but it's really the component that does the job...

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
Jasopaum: Would it be better with ifs and elses?
I have these 3 cases so I don't think I have many alternatives...

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
simonvadee: no, in the StaticValueForm folder directly. something like `<IdentifierSystemInput>`

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
simonvadee: `attribute.definition.id === 'Reference.type' && renderMyComponent()`
That should hold in one line and is more readable in my opinion

Jasopaum: If I do this for the 3 cases, it won't be clear that only 1 of them will be rendered, I think

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
Jasopaum: Done

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
Jasopaum: Do you find it clearer now?

client/src/components/mapping/TabColumnPicking/StaticValueForm/IdentifierSystemInput/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const [customSystem, setCustomSystem] = useState(false);
+  const [selectedSource, setSelectedSource] = useState(
+    undefined as Source | undefined
+  );
+  const [selectedResource, setSelectedResource] = useState(
+    undefined as Resource | undefined
+  );
+
+  // If the current attribute is the identifier of the resource and not a Reference identifier
+  const isIdentifierSystem = /^identifier\[\d+\]\.system$/.test(attribute.path);
```
simonvadee: same comment as before: this belongs in the `Attribute` class

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
simonvadee: yes!

PR #230, opened 2020-03-31
Add modal to manage collaborators

simonvadee:
pyrog-manage-collaborators2

### DIFF
--- a/client/src/components/sources/index.tsx
+++ b/client/src/components/sources/index.tsx
@@ -1,15 +1,6 @@
-import {
-  Alert,
-  Button,
-  Card,
-  Elevation,
-  Icon,
-  Intent,
-  Spinner,
-  Tag
-} from '@blueprintjs/core';
+import { Alert, Button, Intent, Spinner } from '@blueprintjs/core';
 import * as QueryString from 'query-string';
-import React, { ReactElement } from 'react';
+import React from 'react';
 import { useMutation, useQuery } from '@apollo/react-hooks';
 import { useDispatch, useSelector } from 'react-redux';
 import useReactRouter from 'use-react-router';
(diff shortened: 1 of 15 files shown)


server/src/permissions/index.ts
```diff
@@ -65,7 +65,7 @@ export const permissions = shield(
       structureDefinition: rules.isAuthenticatedUser,
     },
     Mutation: {
-      createAccessControl: rules.isAdmin,
+      createAccessControl: rules.isSourceWriter,
```
Jasopaum: Don't you need to add something so that the rule understands the userEmail parameter?

client/src/components/sources/index.tsx
```diff
@@ -40,24 +27,37 @@ const SourcesView = (): React.ReactElement => {
...
 
   const removeSourceFromCache = (
     cache: any,
     { data: { deleteSource } }: any
+  ): void => updateCachedSource(cache, deleteSource, true);
+
+  const updateCachedSource = (
+    cache: any,
+    source: ISelectedSource,
+    toDelete?: boolean
```
Jasopaum: A bit tricky but I understand that it's more compact to have this here.
I'm ok with that if we don't find anything better

server/src/permissions/index.ts
simonvadee: well, the sourceId is still passed in this request and the userId is parsed from the JWT so no need to update this func :)

client/src/components/sources/index.tsx
simonvadee: I prefer doing it here rather than duplicating this code in multiple components. I find it annoying that the apollo cache does not update automatically at every mutation, I think that we could use some plugin that updates the cache automatically instead of having to do it manually. I'll look into it when i'll find the time.

simonvadee:
Here you go ! https://apps.apple.com/us/app/giphy-capture-the-gif-maker/id668208984?mt=12

server/src/permissions/index.ts
Jasopaum: Oh, right

PR #234, opened 2020-04-02
Comments history and authors

simonvadee:
WIP: i'm finishing the import/export features but you can already reveiew the rest since it's quite big.


client/src/components/mapping/Comments/index.tsx
```diff
@@ -22,78 +24,95 @@ const Comments = () => {
...
       attributeId: attributeForNode ? attributeForNode.id : null
     },
     skip: !attributeForNode
   });
 
   React.useEffect(() => {
-    setComments(data && data.attribute.comments ? data.attribute.comments : '');
-  }, [data]);
+    if (attrWithComments)
+      setComments(attrWithComments.attribute.comments.reverse());
```
Jasopaum, line 49: Should we sort by creation time instead?
Not sure but maybe the order of comments could be shuffled on import/export, deletion (when this will be supported), etc.

simonvadee: ok that sounds good

client/src/graphql/fragments/entireAttribute.graphql
```diff
@@ -4,8 +4,16 @@ fragment entireAttribute on Attribute {
   id
   path
   mergingScript
-  comments
+  comments {
```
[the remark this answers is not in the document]
simonvadee: done

client/src/components/mapping/Comments/index.tsx
simonvadee: done

client/src/components/mapping/Comments/index.tsx
```diff
@@ -22,78 +24,130 @@ const Comments = () => {
...
-            path: attribute.path,
-            data: {
-              comments
-            }
+            definitionId: attribute.types[0],
+            path: attribute.path
           }
         });
         const newAttr = data.createAttribute;
+        attributeId = newAttr.id;
```
Jasopaum, line 97: I don't know why this was in the other components but it shouldn't be here!

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -111,7 +113,9 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
       id: attribute.tail,
       secondaryLabel: secondaryLabel(attribute, childNodes),
       nodeData: attribute,
-      label: ''
+      label: '',
+      isSelected:
+        !!selectedAttribute && attribute.path == selectedAttribute.path
```
Jasopaum, line 118: Why do we need that only now?

server/src/resolvers/mapping/utils.ts
```diff
@@ -61,8 +64,15 @@ export const buildInputsQuery = (
     return input
   })
 
-export const buildAttributesQuery = (
-  attributes: AttributeWithInputs[],
+export const buildCommentQueryPreV7 = (
+  comment: string,
+): CommentCreateWithoutAttributeInput => ({
+  content: comment,
+  author: { connect: { email: 'admin@arkhn.com' } },
```
Jasopaum, line 71: What if the admin user is not in the DB?
Would it be worth having user optional for a comment and display something like _anonymous_ when the comment has no identified user?

Jasopaum: This can be done in a later PR if you think it's a good idea but you're sick of working on that!

server/src/resolvers/mapping/utils.ts
```diff
@@ -61,8 +64,15 @@ export const buildInputsQuery = (
...
+})
+
+export const buildAttributesQueryPreV7 = (
```
Jasopaum, line 74: Do you think it would be more readable if we had only one function and pass it also the version?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
simonvadee: we don't need that only now, but i just noticed now that it's disturbing that when you refresh the page, the "selectedAttribute" from the redux store is not selected in the FhirResourceTree.

client/src/components/mapping/Comments/index.tsx
simonvadee: yes it should, I need it in the mutation just below that ?

Jasopaum: Oops, misread!
You don't create the parents in the DB when you add a comment?

server/src/resolvers/mapping/utils.ts
simonvadee: the author of a comment is  not optional so I have to connect the author somehow. I agree that this will break if there is no user "admin@arkhn.com" in the db but it's the easiest way I found to this now. I've discussed it with elsie and nico and we noticed that je issue is bigger than that : when you export a mapping on staging, you want to be able to import it on demo: that will fail if there is no user that matches the author's email. We thought of decorrelating a mapping and its "review" (ie: its comments) but we'll work on that later.

server/src/resolvers/mapping/utils.ts
simonvadee: yes it would, i'll do that. But anyway I think we'll get rid of older versions soon to avoid having to maintain a bunch of versions at the same time

simonvadee: oh no actually I can't because the function signatures are not the same. Before v7, `attribute.comments` is a `string`, after it's a `IComment`

PR #239, opened 2020-04-03
Improve style on comments

LaRiffle:
First PR 🚀 
Capture d’écran 2020-04-04 à 00 42 45

### DIFF
--- a/client/src/components/sources/index.tsx
+++ b/client/src/components/sources/index.tsx
@@ -152,7 +152,7 @@ const SourcesView = (): React.ReactElement => {
             setSourceToDelete(undefined);
           }}
         >
-          <p style={{"color": "#A82A2A"}}>
+          <p className="delete-warning">
             Etes-vous sûr de vouloir supprimer la source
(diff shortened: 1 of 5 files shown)


client/src/components/sources/index.tsx
```diff
@@ -152,10 +152,10 @@ const SourcesView = (): React.ReactElement => {
             setSourceToDelete(undefined);
           }}
         >
-          <p>
-            Etes-vous sûr de vouloir supprimer la source{' '}
-            <b>{sourceToDelete ? sourceToDelete.name : ''}</b>? Cette action
-            n'est pas réversible.
+          <p style={{"color": "#A82A2A"}}>
```
LaRiffle, line 155: What's the clean way to set a color using the standard buleprint js color set?

simonvadee: you should add a class or an id to your HTML element and add styling in the style.css file associated, where you can use "colors variables" directly from blueprint (there is already an example in `.../Comments/style.css`)

LaRiffle: It was in front of my eyes haha
Ok it's done :)

PR #245, opened 2020-04-10
Handle extensions on composiute types

simonvadee:
upload-extension
add-extension

Also I refactore FhirResourceTree for the 100th time and splitted it in two components: TreeNode and FhirResourceTree. @Jasopaum let me know what you think.

Not handled yet: 
- adding extensions on a top-level resource (such as Patient)
- adding extensions on primitive types (eg: string)


client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -115,6 +115,7 @@ const StaticValueForm = ({ attribute }: Props): React.ReactElement => {
   const addStaticValue = async (value: string): Promise<void> => {
     try {
       if (!attributeId) {
+        console.log('create attr', path);
```
Jasopaum: You forgot a log

client/src/components/selects/addExtensionSelect.tsx
```diff
@@ -0,0 +1,149 @@
+import * as React from 'react';
```
Jasopaum, line 1: How different is this file from `addResourceSelect`? Do you think it would be easy to merge these as a generic component? If it's too long, we can put it in a TODO list

client/src/components/uploads/uploadExtension/index.tsx
```diff
@@ -0,0 +1,200 @@
+import React from 'react';
```
Jasopaum, line 1: Same, maybe we can have a generic upload component but definitely not for this PR

server/src/resolvers/StructureDefinition.ts
```diff
@@ -61,12 +62,23 @@ export const StructureDefinition = objectType({
...
+    t.list.field('attributes', {
+      type: 'AttributeDefinition',
       description: 'Structured version of the attributes',
       resolve: (parent: any) => parent.attributes,
     })
 
+    t.list.field('extensions', {
+      type: 'StructureDefinition',
+      description: 'List of extensions on this type',
+      resolve: async (parent: any, { elementId }) => {
```
Jasopaum, line 74: Sorry, I don't get where this `elementId` comes from

server/src/resolvers/StructureDefinition.ts
```diff
@@ -61,12 +62,23 @@ export const StructureDefinition = objectType({
...
-      type: 'JSON',
+    t.list.field('attributes', {
+      type: 'AttributeDefinition',
       description: 'Structured version of the attributes',
       resolve: (parent: any) => parent.attributes,
     })
 
+    t.list.field('extensions', {
+      type: 'StructureDefinition',
+      description: 'List of extensions on this type',
```
Jasopaum, line 73: Maybe it would be clearer to say 'List of allowed/possible extensions on this type.
And I think I'm a bit lost here because I thought the extension depended on the context (which is ~ a path)

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/nodeLabel.tsx
```diff
@@ -2,39 +2,59 @@ import React from 'react';
...
+    <MenuItem
+      icon={'delete'}
+      onClick={deleteNodeCallback}
+      text={"Supprimer l'item"}
+    />
+  );
+
   const showContextMenu = async (e: React.MouseEvent<HTMLDivElement>) => {
     e.preventDefault();
+    const hasExtensions = !isArray && !!extensions && extensions.length > 0;
```
Jasopaum, line 45: Don't you think it's confusing to call it extensions instead of something like allowedExtensions?

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -71,6 +71,7 @@ const StaticValueForm = ({ attribute }: Props): React.ReactElement => {
       setStaticValue('');
       getFhirTypes();
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Jasopaum, line 74: What is this?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -260,23 +251,37 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
-      setSelectedNode(node);
+      setSelectedNode(node as TreeNode);
       onClickCallback(node.nodeData);
     }
     setNodes([...nodes]);
   };
 
+  const handleContextMenu = async (
+    node: ITreeNode<Attribute>
+  ): Promise<void> => {
```
Jasopaum, line 262: Isn't this called when you right-click on a Node? I don't get it

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -260,23 +251,37 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
-        updateSecondaryLabel(node);
-      }
+      nodes.forEach(n => n.updateSecondaryLabel());
       return [...nodes];
     });
   }, [attributesForResource]);
 
   useEffect(() => {
     if (!loading) setNodes(genTreeLevel(fhirStructure));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nriss, line 284: This comment should be removed ?

client/src/components/uploads/uploadProfile/index.tsx
```diff
@@ -116,6 +116,7 @@ const UploadProfile = ({ isOpen, resource, onClose, onUpload }: Props) => {
       }
     };
     reader.readAsText(profileFile);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nriss, line 119: This comment should be removed ?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
simonvadee: nope it can't: it prevents lint errors that we cannot avoid (we cannot have an exhaustive list of dependency array in this specific case).

client/src/components/uploads/uploadProfile/index.tsx
simonvadee: same answer

client/src/components/selects/addExtensionSelect.tsx
simonvadee: it's quite different from addResourceSelect because it not renders profiles but it could be made generic (probably in the same PR as the generic upload)

client/src/components/uploads/uploadExtension/index.tsx
simonvadee: agreed!

server/src/resolvers/StructureDefinition.ts
simonvadee: yeah, we don't need that anymore, sorry I forgot

server/src/resolvers/StructureDefinition.ts
simonvadee: fixed the description. In the context of a StructureDefinition, we don't have any context information (where is the StructureDefinition defined in a fhir object) so we only fetch the extensions that apply on this StructureDefinition in any context. However, when dealing with AttributeDefinition, we DO have context (in the cached attribute definition, the id provides information eg: `Observation.category:VSCat.text`) so we can also fetch all extensions that apply in this specific context

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/nodeLabel.tsx
simonvadee: changed

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
simonvadee: to prevent errors from the linter about the f*cking dependency array on Rect.useEffect(). I don't know since when but now warnings are treated as errors in the CI (yarn does that automatically) so the build couldn't pass otherwise.

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
simonvadee: yes it is, why ?

simonvadee: one is for left-click, the other for right-click

server/src/resolvers/StructureDefinition.ts
Jasopaum: Hum, right, I thought that maybe we could have extensions with a context that is not directly in a single structure definition but it seems that it won't be the case 👍

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
Jasopaum: It looks like you're doing what should be done on a left-click here...

simonvadee: no: the left click should fetch the children and expand the node, the right click should just fetch the nodes without collapsing the node.

Jasopaum: Oh, ok, on left click you generate the sub-tree inside the node so that you can tell which extensions can be applied?

PR #249, opened 2020-04-15
Fix bug ref type dropdown

Jasopaum:
The real problem is a bit deeper: it's the fact that Attribute objects stored in Redux lose their accessors.
This is a temporary fix and we should go back to using accessors when we've found how to do so.


client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -67,7 +67,7 @@ const StaticValueForm = ({ attribute }: Props): React.ReactElement => {
     : null;
 
   useEffect(() => {
-    if (attribute.isReferenceType) {
+    if (attribute.definition.id === 'Reference.type') {
```
simonvadee, line 70: please add an issue or a TODO comment otherwise we're going to forget about it

PR #253, opened 2020-04-17
Clean some components

Jasopaum:
I wanted to remove some of the remaining class components (to replace them with functional components).

- Authentication component is now functional
- Removed an unused component (ColumnViewer)
- What was ColumnPicker is now a function ColumnSelect moved in the selects folder


simonvadee:
tu fais le ménage avant de fermer la boutique quoi, propre

PR #259, opened 2020-04-20
Fix slicing display

### LINKED ISSUE #241: Slicing hierarchy error
When importing a profile with extensions (ex : observation-bp https://www.hl7.org/fhir/bp.html or variant http://hl7.org/fhir/uv/genomics-reporting/variant.html), the slice structure is not as expected.

Example of observation-bp:

* Actual state
image

* Expected
see https://www.hl7.org/fhir/bp.html#profile
image

Jasopaum:
Add difference between slices and choice types

We can choose the type of slice to add in array as follows:
image

Goes with https://github.com/arkhn/fhir.ts/pull/12

Closes #241


client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -195,10 +195,22 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
     const node = createNode(attribute, []);
 
-    if (attribute.types.length > 1) {
+    if (attribute.slices.length > 0) {
+      // if the attribute has slices, we build a node with these different slices inside
+      node.childNodes = buildSlicedNode(attribute);
+    } else if (attribute.types.length > 1) {
       // if the attribute has multiple types, we have to create as many
       // children as there are types.
       node.childNodes = buildMultiTypeNode(attribute);
```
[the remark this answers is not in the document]
Jasopaum: Oh, that's right! And now, I got your point on the spreadType in the other PR
I'll see which one is the best to keep

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -195,10 +195,22 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
       .map(buildNodeFromAttribute)
       .filter(Boolean) as TreeNode[];
 
+  const buildSlicedNode = (attribute: Attribute): TreeNode[] => {
+    const children = attribute.slices.map(buildNodeFromAttribute)
+
+    const copy = Attribute.from({ ...attribute, slices: [] });
+    const base = buildNodeFromAttribute(copy);
```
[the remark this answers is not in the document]
Jasopaum: Yes, I agree that it's a bit odd
It's because slices are stored as "children" of a default slice.
For instance, I can have an attribute and then slices that are specifications of it.
And what I want to show is something like
Capture d’écran 2020-04-21 à 10 58 32

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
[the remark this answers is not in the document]
Jasopaum: Hum, we don't always fill the choices because sometimes, in the structure def, only attribute[x] will be present and not attribute[x].field.
Are you ok if we first check if there are choices and if not, we check for the multi types case?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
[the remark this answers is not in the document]
simonvadee: gotcha! please add a comment to explain

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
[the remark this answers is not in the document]
simonvadee: I think we can't keep both (too confusing), can't we just populate choices automatically when there are multiple types on a choice attribute ? ie: when adding a value[x] attribute, we call attribute.addChoice automatically for each type ?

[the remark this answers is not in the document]
Jasopaum: The problem would be when we have value[x] and then a element with value[x].field in the structure definition... We don't want to add the choice automatically if it's specified after

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/nodeLabel.tsx
```diff
@@ -3,25 +3,55 @@ import { Menu, MenuItem, ContextMenu, Tooltip } from '@blueprintjs/core';
...
+  addSliceCallback,
   addExtensionCallback,
   deleteNodeCallback
-}: NodeLabelProps) => {
+}: NodeLabelProps): React.ReactElement => {
   const client = useApolloClient();
-  const renderAddItem = () => (
-    <MenuItem icon={'add'} onClick={addNodeCallback} text={'Ajouter un item'} />
-  );
+  const renderAddItem = () => {
```
simonvadee, line 34: I think it's best to have a function renderAddSlice and you do the conditional rendering in "showContextMenu"

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/nodeLabel.tsx
```diff
@@ -3,25 +3,55 @@ import { Menu, MenuItem, ContextMenu, Tooltip } from '@blueprintjs/core';
...
+          text={'Ajouter un item'}
+        />
+      );
+    } else {
+      const sliceNames = slices
+        .map(s => s.definition.sliceName)
+        .filter(Boolean) as string[];
+      return (
+        <AddSliceSelect
+          items={['default', ...sliceNames]}
```
simonvadee, line 49: don't use "default"  as discussed

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -143,7 +144,8 @@ const StaticValueForm = ({ attribute }: Props): React.ReactElement => {
             variables: {
               resourceId: resource.id,
               definitionId: currentAttribute.types[0],
-              path: parentPath
+              path: parentPath,
+              sliceName: currentAttribute.definition.sliceName
```
simonvadee, line 148: lache un com!

client/src/components/mapping/TabColumnPicking/DynamicColumnPicker/index.tsx
```diff
@@ -113,7 +114,8 @@ const DynamicColumnPicker = ({ attribute, schema, source }: Props) => {
             variables: {
               resourceId: resource.id,
               definitionId: currentAttribute.types[0],
-              path: parentPath
+              path: parentPath,
+              sliceName: currentAttribute.definition.sliceName
```
simonvadee, line 118: lache un com!

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -219,6 +225,7 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
       node.childNodes = buildSlicedNode(node);
     } else if (attribute.choices.length > 0) {
       // if the node has choices, create a node for each of them
+      attribute.choices.forEach(choice => attribute.parent?.addChild(choice));
```
simonvadee, line 228: I think this is the wrong place to do this. I get the issue: when we do Attribute.from avec having fetched the definition, we don't have access to the parent so `addChoice` doesn't properly set the parent. I think we should provide the parent to the `buildAttributes` function so that Attribute.from() does his job. we can talk about it, maybe i'm getting this wrong

Jasopaum: Yes, that was the problem
I tried to give an optional parent to `buildAttributes` but it didn't work properly not sure why. I can try again tomorrow
Also, this forEach is used just below in `handleContextMenu` so if we find a better way, we'll clean this at several places :)

simonvadee: ok let's try again together tomorrow!

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -247,14 +247,15 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
     attributes.map(buildNodeFromAttribute).filter(Boolean) as TreeNode[];
 
   const fetchAttributeDefinition = async (
-    definitionId: string
+    definitionId: string,
+    parent?: Attribute
```
simonvadee, line 251: should this be optional ?

simonvadee: do you think it's best to pass only the parent, and use parent.types[0] as the definitionId ?

Jasopaum: You're right! I can do that 👍

Jasopaum: Done

PR #261, opened 2020-04-23
Handle extensions on root types

simonvadee:


server/src/resolvers/StructureDefinition.ts
```diff
@@ -71,7 +71,15 @@ export const StructureDefinition = objectType({
     t.list.field('extensions', {
       type: 'StructureDefinition',
       description: 'List of allowed extensions on this type',
-      resolve: async (parent: any) => typeExtensions(parent.meta.id),
+      resolve: async (parent: any) =>
+        // the the definition is not a root type (eg: observation-bodyweight), we need to return
+        // extensions on observation-bodyweight (definition.id) andq Observation (definition.type).
```
Jasopaum, line 76: typo: "andq"

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -117,6 +120,7 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
     node: TreeNode,
     extensionDefinition: IStructureDefinition
   ): void => {
+    console.log('add', extensionDefinition);
```
Jasopaum, line 123: We can remove that, I think

PR #263, opened 2020-04-24
Improve comments and validation

simonvadee:
Screenshot 2020-04-24 at 17 36 11

I changed the SecondaryLabel of a TreeNode (I made an other component). First improvement is: it works better (like refresh and stuff). Otherwise
- green tick when attribute is "validated"
- red dot when the attribute is required
- orange dot when there is an input but attr is not validated
- grey dot when there are children with inputs

I also added the possibility to "validate" an attribute with a big nice button on top of comments. You can't validate an attribute twice (per person)


client/src/components/mapping/FhirMappingPanel/FhirResourceTree/secondaryLabel.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const attributesWithInputs = Object.keys(attributesForResource).filter(
+    path => attributesForResource[path].inputs.length > 0
+  );
+
+  const checkHasChildInputs = (pathString: string) =>
+    attributesWithInputs.some(
+      el => el !== pathString && el.startsWith(pathString)
+    );
+
+  const checkHasInputs = (pathString: string) =>
```
Jasopaum, line 27: Maybe `hasChildInputs` and `hasInputs` would be better than `checkHasChildInputs` and `checkHasInputs`

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/secondaryLabel.tsx
```diff
@@ -0,0 +1,67 @@
...
+    attributesWithInputs.includes(pathString);
+
+  const isValidated = attributesForResource[attribute.path]?.comments.some(
+    c => c.validation
+  );
+  const comments = attributesForResource[attribute.path]?.comments.filter(
+    c => !c.validation
```
Jasopaum, line 34: I think it could be nice to be able to add a comment with a validation but maybe it changed to much the  logic you put in place

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/secondaryLabel.tsx
simonvadee: we already have variables with these names, which are initialised using these functions.

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/secondaryLabel.tsx
simonvadee: yeah I thought about it and it's entirely compatible with the current datamodel (a validation comment may also have a `content`, but right now it's empty) but we will add that when we improve UX later it's that's alright with you

PR #272, opened 2020-04-27
Improve display of validation button

LaRiffle:
I stopped in the middle of the work (with the undo validate button) but here is the first step
Capture d’écran 2020-04-28 à 09 45 43


Jasopaum:
People seem to like that big ol' button

PR #273, opened 2020-04-28
Build and publish pyrog-client for every environment in github actions

### LINKED ISSUE #271: Automatize client image build and publish for all environments

elsiehoffet-94:
Closes #271


client/scripts/build_hospital_client.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
simonvadee, line 1: I would call this script `generate_client_env.sh` (because it does not build the client). Also, maybe we'll have to handle https someday so we'll have to make this script compatible (later).

PR #275, opened 2020-04-28
Enable concept map with ValueSet as target

Jasopaum:


Jasopaum:
> lgtm, you're such a pro in react now I got nothing to say. Do you want to take the lead of the front-end team ?

:man-gesturing-ok:

PR #279, opened 2020-05-04
Add custom export

Jasopaum:
We can choose to include comments or not
And we can also choose to export a bundle containing the concept maps and the structure definitions of profiles used in the mapping

### DIFF
--- a/server/src/resolvers/Query.ts
+++ b/server/src/resolvers/Query.ts
@@ -3,7 +3,7 @@ import { arg, idArg, queryType } from '@nexus/schema'
 import { getDefinition } from 'fhir'
 import { sources } from './Source'
 import { searchConceptMaps } from './ConceptMap'
-import { searchDefinitions } from './StructureDefinition'
+import { searchDefinitions, searchProfiles } from './StructureDefinition'
 
 export const Query = queryType({
   definition(t) {
@@ -102,7 +102,7 @@ export const Query = queryType({
       resolve: searchDefinitions,
     })
 
-    t.list.field('conceptMapIds', {
+    t.list.field('usedConceptMapIds', {
       type: 'String',
       args: {
         sourceId: idArg({ nullable: false }),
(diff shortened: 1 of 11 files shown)


server/src/resolvers/Source.ts
```diff
@@ -183,3 +185,77 @@ export const deleteSource: FieldResolver<'Mutation', 'deleteSource'> = async (
...
+  if (!source) {
+    throw new Error(`source ${sourceId} does not exist`)
+  }
+
+  const resources = await ctx.prisma.resource.findMany({
+    where: { source: { id: sourceId } },
+  })
+
+  const definitions = await Promise.all(
+    resources.map(r => getDefinition(r.definitionId)),
```
simonvadee, line 252: what about profiles on attributes ?

server/src/resolvers/Source.ts
```diff
@@ -183,3 +185,77 @@ export const deleteSource: FieldResolver<'Mutation', 'deleteSource'> = async (
...
+  )
+
+  const profileIds = definitions
+    .filter(def => !!def && def.meta.id !== def.meta.type)
```
simonvadee, line 256: I think it's be best to use the `definition.meta.derivation` which equals to "specialization" if it's a profile and "constraint" if it's a base resource. also, can you explain the case where def is `null` (why do you need !!def ?)

server/src/resolvers/Query.ts
```diff
@@ -100,5 +100,21 @@ export const Query = queryType({
       nullable: true,
       resolve: searchDefinitions,
     })
+
+    t.list.field('usedConceptMapIds', {
```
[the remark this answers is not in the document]
Jasopaum: Hum, at first, I think it was too much to put on Source but now, I agree, it can be there.
I'll change that

server/src/resolvers/Source.ts
Jasopaum: You mean extensions? I don't think I follow

server/src/resolvers/Source.ts
Jasopaum: I think the !! was there because of typing (the find above returns a list of defs | undefined)

Jasopaum: Agreed for the profile check, I'll change that

PR #285, opened 2020-05-06
Raise an error if imported mapping contains unknown authors

Jasopaum:


client/src/components/newSource/index.tsx
```diff
@@ -154,15 +154,23 @@ const NewSourceView = (): React.ReactElement => {
...
-          await client.mutate({
-            mutation: mCreateSource,
-            variables: {
-              templateName,
-              hasOwner,
-              mapping,
-              name: sourceName
-            }
-          });
+          await client
```
simonvadee, line 157: why not using try {} catch (err) instead of the old-style .catch ?

server/src/resolvers/mapping/index.ts
```diff
@@ -47,7 +47,10 @@ export const importMapping = async (
     case MAPPING_VERSION_6:
       return handleV6(prismaClient, sourceId, resources)
     case MAPPING_VERSION_7:
-      return handleV7(prismaClient, sourceId, resources)
+      const existingUsers = (await prismaClient.user.findMany()).map(
```
simonvadee, line 50: that's a little bit barbaric, I was thinking of doing the query like this:
```
const authors = comments.map(c => c.author.email).filter(filterDuplicates)
const res = prisma.user.findMany({email: {in: authors}})
const missingUsers = authors.map(a => !res.includes(a))
if (missingUsers.length > 0) throw Error(`missing users ${missingUsers.join(', '}`)
```
which would avoid fetching all users at once

client/src/components/newSource/index.tsx
Jasopaum: I'm not sure exactly how the FileReader and this function work together, I broke that once... So I went for the minimal change 😅
But I can try again quickly

server/src/resolvers/mapping/utils.ts
```diff
@@ -25,6 +26,35 @@ export const clean = (entry: any): any => {
...
+        (resourceAuthors: string[], attribute: AttributeWithComments) => [
+          ...resourceAuthors,
+          ...attribute.comments.map(comment => comment.author.email),
+        ],
+        [],
+      ),
+    ],
+    [],
+  )
+  const uniqueMails = authorMails.filter(
```
simonvadee, line 46: rename to uniqueEmails

server/src/resolvers/mapping/utils.ts
```diff
@@ -25,6 +26,35 @@ export const clean = (entry: any): any => {
...
+    (mail, ind) => authorMails.indexOf(mail) === ind,
+  )
+  const allUsers = await prismaClient.user.findMany({
+    where: { email: { in: uniqueMails } },
+  })
+  const existingMails = allUsers.map(user => user.email)
+  const missingUsers = uniqueMails.filter(mail => !existingMails.includes(mail))
+  if (missingUsers.length > 0)
+    throw Error(`missing users ${missingUsers.join(', ')}`)
```
simonvadee, line 55: I liked your previous error message "trying to import a mapping with unexisting comment author ${missingUsers.join(', '}"

PR #290, opened 2020-05-11
Some bugfixes

Jasopaum:
- fix await on createSource
- enable import mapping without comments
- delete comments and columns on cascade when deleting source


server/src/resolvers/Source.ts
```diff
@@ -182,11 +184,17 @@ export const deleteSource: FieldResolver<'Mutation', 'deleteSource'> = async (
                       where: { id: j.id },
                     })
                   }),
-                )
+                  ctx.prisma.column.delete({ where: { id: i.sqlValue.id } }),
```
simonvadee, line 187: thats better

PR #292, opened 2020-05-14
Add source id when importing/exporting mapping

Jasopaum:


server/src/resolvers/Source.ts
```diff
@@ -87,8 +87,15 @@ export const createSource: FieldResolver<'Mutation', 'createSource'> = async (
...
   // create the source
+  let sourceId
+  let parsedMapping
+  if (mapping) {
+    parsedMapping = JSON.parse(mapping)
+    sourceId = parsedMapping.source.id
+  }
   const source = await ctx.prisma.source.create({
     data: {
+      id: sourceId,
```
simonvadee, line 98: parsedMapping?. ...

server/src/resolvers/mapping/index.ts
```diff
@@ -24,32 +24,31 @@ import { getDefinition } from 'fhir'
 export const importMapping = async (
   prismaClient: PrismaClient,
   sourceId: string,
-  mapping: string,
+  mapping: any,
 ) => {
-  const { version, resources } = JSON.parse(mapping)
-  if (!version) {
+  if (!mapping.version) {
```
simonvadee, line 29: const {resources, version } = mapping
if (!version)
...

server/src/resolvers/mapping/index.ts
```diff
@@ -24,32 +24,31 @@ import { getDefinition } from 'fhir'
 export const importMapping = async (
```
simonvadee, line 24: this is used in src/scripts/importMapping

PR #302, opened 2020-06-03
Use same key and algorithm as auth to generate tokens

Jasopaum:


client/src/app.tsx
```diff
@@ -82,6 +83,12 @@ const store = finalCreateStore(persistedReducer);
 
 const persistor = persistStore(store);
 
+// AXIOS
+
+// Set a default authentication header for fhir api calls
+const token = localStorage.getItem(AUTH_TOKEN);
+axios.defaults.headers.common['Authorization'] = `Bearer ${token}`;
```
simonvadee, line 90: use interceptor instead

PR #326, opened 2020-07-02
Limit number of elements in stringSelect

Jasopaum:


client/src/components/selects/stringSelect.tsx
```diff
@@ -27,8 +27,10 @@ export default class StringSelect extends React.Component<ISelectProps, any> {
     return <MenuItem key={item} onClick={handleClick} text={item} />;
   };
 
-  private filterByName: ItemPredicate<string> = (query, item) => {
-    return `${item.toLowerCase()}`.indexOf(query.toLowerCase()) >= 0;
+  private filterList: ItemListPredicate<string> = (query, items) => {
+    return items
+      .filter(item => `${item.toLowerCase()}`.indexOf(query.toLowerCase()) >= 0)
+      .slice(0, 100);
```
simonvadee, line 33: can we still search through all the items in the list ? and we see the 100 first results ? Maybe this should be a props to make this value configurable ?

Jasopaum: Yep, we can search through the whole list and only 100 items are displayed. We could, I'll add that 👍

PR #329, opened 2020-07-03
Clean GH workflow

Jasopaum:
Not sure it's the cleanest nor if it even works. Maybe it would be better to have a plain script to do that?

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -74,10 +74,8 @@ jobs:
           if ${{ matrix.environment.secret }} != ""; then
             cd client
             ./scripts/generate_client_env.sh > .env.${{ matrix.environment.name }}
-            docker build -t arkhn/pyrog-client:${{ matrix.environment.name }} --build-arg ENV_FILE=.env.${{ matrix.environment.name }} .
-          else
-            yarn build:docker:${{ matrix.environment.name }}
           fi
+          docker build -t arkhn/pyrog-client:${{ matrix.environment.name }} --build-arg ENV_FILE=.env.${{ matrix.environment.name }} .
 
       - name: Tag docker image with hash
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/client.yml
```diff
@@ -49,76 +49,37 @@ jobs:
...
-          cd client
-          ./scripts/generate_client_env.sh > .env.curie
-          docker build -t arkhn/pyrog-client:curie --build-arg ENV_FILE=.env.curie .
-
-      - name: Build RMS
+      - name: Build image
         env:
-          IP: ${{ secrets.IP_VM_REIMS }}
-          STAGE: rms
+          IP: ${{ secrets[matrix.environment.secret] }}
```
elsiehoffet-94, line 70: Isn't this "" for demo, staging and dev?

.github/workflows/client.yml
```diff
@@ -49,76 +49,37 @@ jobs:
...
+          ]
     steps:
       - uses: actions/checkout@v1
 
-      - name: Build staging and demo
-        run: |
-          cd client/
-          yarn build:docker:staging
-          yarn build:docker:dev
-          yarn build:docker:demo
```
elsiehoffet-94, line 60: Shouldn't we keep those? or add IP_STAGING and IP_DEMO in the strategy?

Jasopaum: Oops, indeed I need to add IP_staging, etc. in the GH secrets if I want to do that
Which makes me even less convinced that this is the best solution 😅

simonvadee: good catch @elsiehoffet-94. I don't think it's that bad to add the url of our environments in GH secrets BUT if you're not convinced there is an other solution:
- keep staging,demo,dev in the matrix and remove the "secret" field (or keep in empty)
- do not generate a .env.<target> file for environments which do not have the "secret" field in the matrix --> use the .env.[dev|staging|demo] that we push in the repo instead

.github/workflows/client.yml
simonvadee: yup, this wont work as is

simonvadee:
I think it's pretty clever/clear to use the `matrix` feature of GH actions + a script using an env var (IP_ENV). I'm not sure if a plain script would make it clearer.

PR #343, opened 2020-07-31
Enable conditional attributes

Jasopaum:


client/src/components/mapping/TabColumnPicking/DynamicColumnPicker/index.tsx
```diff
@@ -35,31 +38,42 @@ const DynamicColumnPicker = ({ attribute, schema, source }: Props) => {
...
+  );
   const attributesForResource = useSelector(
     (state: IReduxStore) => state.resourceInputs.attributesMap
   );
+
+  const path = attribute.path;
   let attributeId = attributesForResource[path]
     ? attributesForResource[path].id
     : null;
+  let inputGroupId =
```
simonvadee, line 52: please explain wtf is going on here, this is too hard to read without a pint of coffee

client/src/components/mapping/TabColumnPicking/StaticValueForm/index.tsx
```diff
@@ -62,6 +68,12 @@ const StaticValueForm = ({ attribute }: Props): React.ReactElement => {
   let attributeId = attributesForResource[path]
     ? attributesForResource[path].id
     : null;
+  let inputGroupId =
```
simonvadee, line 71: same

client/src/graphql/mutations/addConditionToInputGroup.graphql
```diff
@@ -0,0 +1,22 @@
+#import "../fragments/entireCondition.graphql"
+
+mutation addConditionToInputGroup(
+  $inputGroupId: ID!
+  $action: String
```
simonvadee, line 5: isn't this required ? as well as the other fields ?

client/src/services/selectedNode/reducer.tsx
```diff
@@ -3,7 +3,8 @@ import { ISimpleAction } from 'types';
 const initialState: any = {
   source: null,
   resource: null,
-  attribute: null
+  attribute: null,
+  selectedInputGroup: 0
```
simonvadee, line 7: i don't think it's a good idea to initialize this to 0, I would use null instead. Would that cause an issue elsewhere ?

client/src/graphql/mutations/addConditionToInputGroup.graphql
Jasopaum: I first create empty conditions and fill them after. So we could even remove these arguments for the addConditionToInputGroup mutation. What do you think?

client/src/components/mapping/InputCondition/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+      >
+        <Tag minimal={true}>VALUE</Tag>
+        <Tag intent={'primary'} large={true}>
+          <input
+            className="text-input"
+            value={value}
+            type="text"
+            placeholder="value..."
+            onChange={(e: React.ChangeEvent<HTMLInputElement>): void => {
+              // NOTE value updated in DB at each character. Put a submit button?
```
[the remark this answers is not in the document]
Jasopaum: I just made an issue for that (https://github.com/arkhn/pyrog/issues/347)

server/prisma/migrations/migrate.lock
```diff
@@ -1,4 +1,5 @@
 # Prisma Migrate lockfile v1
 
 20200623144855-database-schema
-20200803162228-input-groups
\ No newline at end of file
+20200803162228-input-groups
```
simonvadee, line 4: duplicate

PR #349, opened 2020-08-06
POC of d3 fhir mapping graph

simonvadee:


d3-fhir-graph/src/index.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import './index.css';
+import App from './App';
+import * as serviceWorker from './serviceWorker';
```
Jasopaum, line 5: I think we can remove this

d3-fhir-graph/src/App.test.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
Jasopaum, line 1: We don't need test files for React apps here

d3-fhir-graph/src/App.js
```diff
@@ -0,0 +1,624 @@
...
+    ];
+  }, []);
+  console.log(nodes, links);
+  return { nodes, links };
+};
+
+// parse(mapping);
+const data = parse(mapping);
+
+// const data = {
```
Jasopaum, line 65: Can we remove the mock data?

d3-fhir-graph/src/index.js
simonvadee: done

PR #350, opened 2020-08-06
Update identifiers systems on import

simonvadee:
Three changes in this PR:
- the way we format identifiers' system `http://terminology.arkhn.com/<sourceId>/<resourceId>[/<optionalCustomKey>` --> `http://terminology.arkhn.com/<resourceId>[/<optionalCustomKey>`

- remove IDs when importing mappings: no more issues with duplicates

- update the identifiers system when importing a mapping in order to match the new resourceID


server/prisma/migrations/20200807160353-resource-logical-reference/migrate-identifiers-system.ts
```diff
@@ -0,0 +1,112 @@
...
+        a.inputGroups.length &&
+        a.inputGroups[0].inputs &&
+        a.inputGroups[0].inputs.length &&
+        a.inputGroups[0].inputs[0].staticValue
+      if (!hasInput)
+        console.warn(`[${a.resource!.definitionId}.${a.path}] missing input`)
+      return hasInput
+    })
+    .map(a => {
+      let prevSystemValue = a.inputGroups[0].inputs[0].staticValue!
```
Jasopaum, line 57: You're only updating the first input of the first input group right?

PR #355, opened 2020-08-11
Change the way to seed credentials

Jasopaum:


server/scripts/importMapping.ts
```diff
@@ -54,11 +84,19 @@ const main = async (path: string) => {
       .map((r: any) => r.definitionId)
       .join(', ')}...`,
   )
-  return importMapping(prismaClient, source.id, mapping)
+
+  let ret: Promise<any> = importMapping(prismaClient, source.id, mapping)
+
+  if (process.argv.length === 4) {
+    console.log(`Seeding credentials from ${process.argv[3]}...`)
+    ret = Promise.all([ret, importCredentials(process.argv[3], source.id)])
```
simonvadee, line 92: you could use prismaClient.$transaction here, right ?

PR #364, opened 2020-08-14
Enhance conditions

Jasopaum:


client/src/components/mapping/InputCondition/index.tsx
```diff
@@ -30,9 +30,22 @@ interface Props {
 }
 
 const availableActions = ['INCLUDE', 'EXCLUDE'];
+const conditionRelations = ['EQ', 'LT', 'LE', 'GE', 'GT', 'NULL', 'NOTNULL'];
```
simonvadee, line 33: you could use Object.keys(conditionToString) for this (and maybe rename conditionToString to `availableConditions` or `allowedConditions` or `conditionsMap` or something else)

PR #379, opened 2020-09-15
Use token from oauth2 token provider to fetch structure definitions

Jasopaum:
Do not merge. Some work needs to be done elsewhere


simonvadee (approved):
does this mean we have to deploy hydra and identity-provider for this to work ?

Jasopaum:
> does this mean we have to deploy hydra and identity-provider for this to work ?

Yes, we need to wait a bit before merging this one 🙂

PR #382, opened 2020-09-17
Use OIDC provider to get tokens

Jasopaum:


client/src/app.tsx
```diff
@@ -124,11 +154,38 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
   if (networkError) console.log(`[Network error]: ${networkError}`);
 });
 
+const afterwareLink = onError(({ graphQLErrors, operation, forward }) => {
+  if (graphQLErrors) {
+    for (const err of graphQLErrors) {
+      if (err.message.includes('token is invalid')) {
```
simonvadee, line 160: can't we have a ststusCode instead ?

client/src/components/routes/privateRoute/index.tsx
```diff
@@ -1,22 +1,83 @@
...
+    axios.interceptors.request.use(config => {
+      config.headers.Authorization = `Bearer ${accessToken}`;
+      return config;
+    });
+
+    // Get user info
+    const decodedIdToken: any = jwt_decode(idToken!);
+
+    // Upsert user from IdP into Pyrog DB
+    upsertUser({
```
simonvadee, line 63: is it the client's responsibility to do that ? I find it safer if it's the serve that updates the user info based on the ID token rather than the client telling pyrog-server to update the user with potentialy arbitrary data

server/prisma/schema.prisma
```diff
@@ -193,7 +193,6 @@ model User {
   id            String          @default(cuid()) @id
   email         String          @unique
   name          String
-  password      String
```
simonvadee, line 196: we will need to migrate all users to identity-provider when applying this migration where the stack is already deployed, will we do this manually ?

server/src/permissions/index.ts
```diff
@@ -66,6 +77,8 @@ export const permissions = shield(
       structureDefinition: rules.isAuthenticatedUser,
     },
     Mutation: {
+      upsertUser: rules.isAuthenticatedUser, // TODO fix so that user can only update itself?
```
simonvadee, line 80: you forgot logout ?

client/src/app.tsx
```diff
@@ -85,12 +87,37 @@ const persistor = persistStore(store);
...
-// Interceptor to add authorization header for each requests
-axios.interceptors.request.use(config => {
-  const token = localStorage.getItem(AUTH_TOKEN);
-  config.headers.Authorization = `Bearer ${token}`;
-  return config;
-});
+// Add an interceptor to refresh access token when needed
+// NOTE we currently only need that for calls to the fhir api, not
+// for gql calls to pyrog-server because pyrog-server caches the user info.
+// Is this what we want? Do we want expiring users in Pyrog too?
```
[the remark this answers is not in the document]
Jasopaum: This was already done, I forgot to remove this comment

client/src/app.tsx
Jasopaum: I wish! But I didn't find out how... Do you know how we could do that?

client/src/components/routes/privateRoute/index.tsx
Jasopaum: I was thinking the same thing. Do you think I should totally remove the upsertUser mutation and upsert at each token introspection in the server (in the getUser function in utils.ts)?

Jasopaum: I left a comment about that, I think

server/prisma/schema.prisma
Jasopaum: I think we'll need a script to do that

client/src/components/authentication/index.tsx
```diff
@@ -1,254 +1,43 @@
...
   );
 
   return (
     <div>
       <Navbar />
-      <div className="formulaires">
-        {signupForm}
-        {loginForm}
-      </div>
+      <div className="formulaires">{loginForm}</div>
```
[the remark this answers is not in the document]
Jasopaum: It will be, the route is already there but not the UI

server/src/server.ts
```diff
@@ -29,6 +33,21 @@ const options: Options = {
     allowedHeaders: ['Authorization', 'Content-Type'],
   },
   bodyParserOptions: { limit: '10mb', type: 'application/json' },
+  formatError: (err: any) => {
```
simonvadee, line 36: booyah

simonvadee (approved):

PR #386, opened 2020-09-21
fix: json parse error on Sources component and TypeError in Mapping component

ghost:
* Clicking on a source sometimes does not work, the page has to be refreshed -> JSON parse error in Sources component
* Refreshing on a ressource page will lead to a bug (dark page) -> TypeError in Mapping component

### DIFF
--- a/client/src/components/sources/index.tsx
+++ b/client/src/components/sources/index.tsx
@@ -71,9 +71,7 @@ const SourcesView = (): React.ReactElement => {
     { update: removeSourceFromCache, onError: onError(toaster) }
   );
 
-  const onSelectSource = async (source: ISelectedSource) => {
-    if (source.credential && typeof source.credential.schema === 'string')
-      source.credential.schema = JSON.parse(source.credential.schema);
+  const onSelectSource = async (source: ISelectedSource): Promise<void> => {
     dispatch(changeSelectedSource(source));
     history.push({
       pathname: '/mapping',
--- a/client/src/components/sources/index.tsx
+++ b/client/src/components/sources/index.tsx
@@ -72,14 +72,8 @@ const SourcesView = (): React.ReactElement => {
   );
 
   const onSelectSource = async (source: ISelectedSource) => {
-    try {
-      if (source.credential)
-        source.credential.schema = JSON.parse(
-          source.credential.schema as string
-        );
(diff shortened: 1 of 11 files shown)


client/src/components/sources/index.tsx
```diff
@@ -66,14 +61,25 @@ const SourcesView = (): React.ReactElement => {
...
   const onSelectSource = async (source: ISelectedSource) => {
-    if (source.credential)
-      source.credential.schema = JSON.parse(source.credential.schema as string);
+    try {
+      if (source.credential)
+        source.credential.schema = JSON.parse(
+          source.credential.schema as string
+        );
+    } catch {
+      // No parsing needed as `schema` attribute is already a valid js object
```
Jasopaum, line 81: Do you when it's a string and not an object?

client/src/components/sources/sourceCard.tsx
```diff
@@ -88,6 +88,6 @@ export const SourceCard = ({
         onClose={() => setIsCollaboratorDialogOpen(false)}
         onUpdate={onUpdate}
       />
-    </React.Fragment>
+    </>
```
Jasopaum, line 91: Why did you remove that? I don't think I have any preference but this and `React.Fragment` are equivalent

client/src/components/sources/index.tsx
```diff
@@ -66,14 +61,19 @@ const SourcesView = (): React.ReactElement => {
...
+
   const [deleteSource, { loading: deletingSource }] = useMutation(
     mDeleteSource,
     { update: removeSourceFromCache, onError: onError(toaster) }
   );
 
   const onSelectSource = async (source: ISelectedSource) => {
-    if (source.credential)
-      source.credential.schema = JSON.parse(source.credential.schema as string);
+    if (source.credential && typeof source.credential.schema === 'string')
```
simonvadee, line 75: I thought the JSON.parse error was thrown when source.credential.schema is an empty string, isn't it ?
I think you can just do
```javascript
if (source.credential?.schema)
...
```
no ?

simonvadee: ok I misunderstood the issue, let's put this in the reducer

simonvadee (approved):

PR #387, opened 2020-09-21
fix: string sort in column picker

ghost:


client/src/components/selects/stringSelect.tsx
```diff
@@ -33,25 +33,19 @@ export default class StringSelect extends React.Component<ISelectProps, any> {
...
-    });
-    return resources;
-  };
+  private sortItems = (resources: string[]): string[] =>
+    resources.sort((s1: string, s2: string): number => s1.localeCompare(s2));
 
-  private displayItem = function(item: string): string {
-    return item ? item : 'None';
-  };
+  private displayItem = (item: string): string => (item ? item : 'None');
```
Jasopaum, line 48: Does `item || 'None'` work? I find this clearer

client/src/components/selects/stringSelect.tsx
```diff
@@ -33,25 +33,19 @@ export default class StringSelect extends React.Component<ISelectProps, any> {
...
-      const name1 = s1.toLowerCase();
-      const name2 = s2.toLowerCase();
-      if (name1 < name2) return -1;
-      if (name1 > name2) return 1;
-      return 0;
-    });
-    return resources;
-  };
+  private sortItems = (resources: string[]): string[] =>
+    resources.sort((s1: string, s2: string): number => s1.localeCompare(s2));
```
Jasopaum, line 46: 👍

client/src/components/selects/stringSelect.tsx
```diff
@@ -33,25 +33,19 @@ export default class StringSelect extends React.Component<ISelectProps, any> {
   };
 
   private filterList: ItemListPredicate<string> = (query, items) => {
-    return items
-      .filter(item => `${item.toLowerCase()}`.indexOf(query.toLowerCase()) >= 0)
-      .slice(0, this.props.maxItems || 100);
+    const stringSet = new Set(
+      items.filter(item => item.toLowerCase().startsWith(query.toLowerCase()))
+    );
```
Jasopaum, line 38: The objective is to add the items starting with the query first, right?
Does it solve our problem if a lot of items start with the string we want to find in the list?

client/src/components/selects/stringSelect.tsx
simonvadee: :+1:

client/src/components/selects/stringSelect.tsx
simonvadee: I think that the improvement to the `sortItem` function below (using string.localeCompare) solves this problem: the closest match should appear first.
Also, since the array is not sliced anymore, all matches will appear in the list (even if it's very long). To me, this is the behaviour we expect.

simonvadee (approved):

PR #390, opened 2020-09-24
fix: bad request when editing a resource

### LINKED ISSUE #371: Improve filters on a ressource
- Add possibility to rename the label of the resource (currently returns a 400)
- Add possibility to modify the filter input, relation, or value (currently returns a 400) 

## Step to reproduce the bug
- Import a new source
- Try to modify a name of a resource containing Filters
- The message "Network error: Response not successful: Received status code 400" appears

tevariou:
## Fixes

Fixes #371 by @elsiehoffet-94 
## Description

* Remove id from sqlColumn field in Filter type definition => doesn't exist in graphql schema
* Remove __typename attr from filters attr before updating a resource => see related issue

## Technical details

Related issue: https://github.com/apollographql/apollo-client/issues/1913

## Tests

1. Select a source
2. Select a resource which has filters
3. Edit the resource
4. Edit the label or the filter and save. It shouldn't return an error 400 anymore and the fields should be correctly updated.


tevariou:
Related issue: https://github.com/apollographql/apollo-client/issues/1913

client/src/graphql/fragments/entireFilter.graphql
```diff
@@ -1,7 +1,5 @@
 fragment entireFilter on Filter {
-  id
   sqlColumn {
-    id
```
Jasopaum, line 4: Do we still need to remove these?

Jasopaum (approved):

simonvadee (commented):
nice!

PR #395, opened 2020-09-28
Skip authorization checks when env=test

Jasopaum:
## Fixes

Fixes #[issue number] by @[issue author]

### DIFF
--- a/server/src/permissions/index.ts
+++ b/server/src/permissions/index.ts
@@ -13,8 +13,12 @@ export const authorizationError = {
   statusCode: 403,
 }
 
+// We want to skip authorization checks during integration tests
+const skipAuthorizations = process.env.ENV === 'test'
+
 const rules = {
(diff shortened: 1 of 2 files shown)


server/src/permissions/index.ts
```diff
@@ -13,8 +13,12 @@ export const authorizationError = {
   statusCode: 403,
 }
 
+// We want to skip authorization checks during integration tests
+const skipAuthorizations = process.env.ENV === 'test'
```
simonvadee, line 17: I would have split this between constants and permissions:
```javascript
# constants.ts
const ENV = process.env.ENV

# permissions.ts
import {ENV} from "constants"
const skipAuthorization = ENV === "test"
const rules = skipAuthorization ? {} : { ...permissions resolvers...}
```

Jasopaum: I hesitated but you're right, it might actually be cleaner to do so

simonvadee (approved):

PR #402, opened 2020-10-01
Set user in redux when access tokens in localstorage

Jasopaum:
## Fixes

Fixes #[issue number] by @[issue author]


simonvadee (approved):
that was easy :)

PR #403, opened 2020-10-01
Ignore authentication during tests

Jasopaum:
## Description
We want to avoid authentication stuff during integration tests.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -12,6 +12,7 @@ import { schema } from './schema'
 import { createContext } from './context'
 import { bootstrapDefinitions } from './fhir'
 import { authClient } from './oauth'
+import { ENV } from './constants'
 
 // AXIOS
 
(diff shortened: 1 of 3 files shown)


server/src/server.ts
```diff
@@ -57,7 +58,9 @@ const options: Options = {
 const { PORT } = process.env
 
 const main = async () => {
-  await setAccessToken()
+  if (!(ENV && ENV === 'test')) {
```
simonvadee, line 61: I think we could use a small refacto here:
```javascript
// constants.ts
const  TEST_ENV = "test"
const IN_PROD = ENV !== TEST_ENV

// server.ts
import { IN_PROD } from './constants'
if (IN_PROD) {
    await setAccessToken()
}

File: server/src/utils.ts [hunk 2]
```diff
@@ -6,12 +6,21 @@ import { Request } from 'express'
...
+  ENV,
+  TOKEN_INTROSPECTION_URL,
+  USER_INFO_URL,
+} from './constants'
 
 export const getUser = async (
   request: Request,
   prisma: PrismaClient,
 ): Promise<User | null> => {
+  if (ENV === 'test')
```
simonvadee, line 20: same

simonvadee (changes requested):

simonvadee (approved):

PR #406, opened 2020-10-02
Refresh pyrog server's token when needed

Jasopaum:
## Fixes
After the token was expired the `refreshDefinition` mutation couldn't be performed anymore.

## Description
I simply added an axios interceptor to ask for a new token when it's expired.


server/src/server.ts
```diff
@@ -24,6 +24,24 @@ const setAccessToken = async () => {
   ] = `Bearer ${fhirApiToken.accessToken}`
 }
 
+// Add an interceptor to ask for another access token when needed
+axios.interceptors.response.use(
+  response => {
```
simonvadee, line 29: don't need brackets

simonvadee (approved):

elsiehoffet-94 (approved):
Seems legit

PR #415, opened 2020-10-14
Pass definition.type as resource type to loader

Jasopaum:
## Fixes
ETL on profiles was not working because we didn't pass the proper resource type in the delete-resource query

## Description
Use the `type` field instead of `id` field in the structure def


client/.env
```diff
@@ -1,14 +1,14 @@
...
 REACT_APP_CLIENT_ID=local-pyrog-client
 REACT_APP_CLIENT_SECRET=pyrog-client
-REACT_APP_AUTH_URL=http://localhost:4444/oauth2/auth
-REACT_APP_TOKEN_URL=http://localhost:4444/oauth2/token
-REACT_APP_REVOKE_URL=http://localhost:4444/oauth2/revoke
-REACT_APP_LOGOUT_URL=http://localhost:4444/oauth2/sessions/logout
+REACT_APP_AUTH_URL=https://staging.arkhn.com/hydra/oauth2/auth
+REACT_APP_TOKEN_URL=https://staging.arkhn.com/hydra/oauth2/token
+REACT_APP_REVOKE_URL=https://staging.arkhn.com/hydra/oauth2/revoke
+REACT_APP_LOGOUT_URL=https://staging.arkhn.com/hydra/oauth2/sessions/logout
```
elsiehoffet-94, line 12: Are you sure we want to commit this ?

elsiehoffet-94 (changes requested):

client/.env
simonvadee: no we should keep local urls, but the .env should be in the gitignore an not be comitted anyway.

simonvadee (approved):
either revert changes on .env or add it to gitignopre and add a line in the README to explain how to use the .env

client/README.md
```diff
@@ -24,15 +24,26 @@ yarn start
...
-ENGINE_BACKEND_URL=https://engine.arkhn.org
-HTTP_BACKEND_URL=https://graphql.pyrog.staging.arkhn.org
-WS_BACKEND_URL=ws://graphql.pyrog.staging.arkhn.org
-CLEANING_SCRIPTS_URL=http://localhost:5000
-```
+`.env.template` contains env variables used by the application.
+To launch the applicaiton with a custom setup, you need to modify it to point to a specific services. This is useful if you are working on the client part of Pyrog only and do not want to run a Pyrog server locally.
+To use it, just save the new file as `.env` and execute `yarn start`.
+
+Here is a brief explanation of the different variables:
```
elsiehoffet-94, line 31: Thanks for this!

elsiehoffet-94 (approved):

simonvadee (approved):

PR #423, opened 2020-10-22
Only display attributes that are filled for readers

### LINKED ISSUE #414: [Data catalog] Change READER permissions
## Problem

The goal is there to get a first version of a data catalog

## Description

When a resource is selected, READER users should only see the mapped attributes
FHIR attributes that are not mapped shouldn't be displayed

## Next steps

The next steps will consist on generating tables and schemes

## Roles
- visualiser: see the mapped attributes + comments
- validator: see the mapped attributes (+ unmapped ?), the comments and can validate each attribute
- writer: can see mapped/unmapped attributes and can change the mapping
- administrator: can do anything

Jasopaum:
## Fixes
Fixes #414 by @nriss 

## Description
Only display attributes that are filled for readers

### DIFF
--- a/client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
+++ b/client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
@@ -39,9 +39,10 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
   const dispatch = useDispatch();
 
   const toaster = useSelector((state: IReduxStore) => state.toaster);
-  const { resource, attribute: selectedAttribute } = useSelector(
+  const { resource, source, attribute: selectedAttribute } = useSelector(
     (state: IReduxStore) => state.selectedNode
   );
+  const user = useSelector((state: IReduxStore) => state.user);
   const baseDefinitionId = resource.definition.id;
 
   const { data, loading } = useQuery(qStructureDisplay, {
@@ -60,6 +61,16 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
   const attributesForResource: { [k: string]: IAttribute } = useSelector(
     (state: IReduxStore) => state.resourceInputs.attributesMap
   );
+  const attributesWithInputs = Object.keys(attributesForResource).filter(
+    path => attributesForResource[path].inputGroups.length > 0
+  );
+
+  const isAdmin = user && user.role === 'ADMIN';
+  const isWriter =
+    source &&
+    source.accessControls.filter(
+      acl => acl.role === 'WRITER' && acl.user.id === user.id
+    ).length > 0;
 
   const buildAttributes = (parent?: Attribute) => ({
     attribute,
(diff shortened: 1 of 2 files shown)


nriss (approved):
Looks good to me!
It will be interesting to have another user as "viewer" that cannot validate mapping, but it can be in another PR

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -60,6 +61,16 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
   const attributesForResource: { [k: string]: IAttribute } = useSelector(
     (state: IReduxStore) => state.resourceInputs.attributesMap
   );
+  const attributesWithInputs = Object.keys(attributesForResource).filter(
+    path => attributesForResource[path].inputGroups.length > 0
```
simonvadee, line 65: may an inputGroup be empty ?

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -60,6 +61,16 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
+  );
+
+  const isAdmin = user && user.role === 'ADMIN';
+  const isWriter =
+    source &&
+    source.accessControls.filter(
+      acl => acl.role === 'WRITER' && acl.user.id === user.id
+    ).length > 0;
```
simonvadee, line 73: would it be useful to do that in the backend ? we would update the `structureDisplay` resolver to only return attributes with inpoutGroups in case the user is a READER. 
I'm not sure it's useful though

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
```diff
@@ -251,8 +262,19 @@ const FhirResourceTree = ({ onClickCallback }: Props) => {
...
   };
 
+  // For source readers, we only want to display attributes that are filled
+  const shoudlBeDisplayed = (pathString: string) =>
+    isAdmin ||
+    isWriter ||
+    attributesWithInputs.some(
+      el => el !== pathString && el.startsWith(pathString)
+    );
+
```
simonvadee, line 272: @elsiehoffet-94  @nriss @Peltier10  do we want to make this switchable ? would it be useful for an ADMIN to benefit form this feature (only see filled attributes) ?

nriss: > @elsiehoffet-94 @nriss @Peltier10 do we want to make this switchable ? would it be useful for an ADMIN to benefit form this feature (only see filled attributes) ?

For now, I'm not sure I will use that kind of functionality as a mapper.
Personally, I prefer seeing all available attributes

elsiehoffet-94: - I think it's an enjoyable option for an admin or writer user to only see filled attributes for a final check of our mapping. 

- I agree with @nriss on the second point, it would be nice to develop the tree on click for a terminal attribute, but this might be complicated for complex arrays, with multiple attributes of type CodeableConcepts. I think it would be prefereable to have this option only on filled attributes. And it's a feature request, and can be the object of a different issue.

elsiehoffet-94 (approved):
🚢 Very cool! 
I don't know if it can be done easily on this PR (probably not), but it would be interesting to enable a general comment on the ressource, and on the source. Kind of like a README file, in markdown, to help describe the data available.

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
Peltier10: I agree with @elsiehoffet-94 and @nriss. It can be useful to make a last check up of the mapping.

Jasopaum:
Features outside the scope of this PR that have been asked for have issues here: #289, #425

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
Jasopaum: After discussion, we stated that this won't be done at the server level

client/src/components/mapping/FhirMappingPanel/FhirResourceTree/index.tsx
Jasopaum: Only if something bad happens... But I still fixed that

PR #436, opened 2020-11-03
Add a updateRole mutation

### LINKED ISSUE #432: Expose a mutation to make a USER an ADMIN
Currently, to do so, we manually change the role in the DB.
Note that only an ADMIN should be able to give admin rights to someone else

Jasopaum:
## Fixes
Fixes #432


simonvadee (approved):
nit: I think updateRole would be more appropriate

PR #440, opened 2020-11-05
Add script to add a superuser

### LINKED ISSUE #434: Add a "create admin" script
- script in pyrog-server
- create user in identity-provider (user may already exist)
- create user in pyrog db
- configurable email/password/role (via env vars or CLI args)

tevariou:
## Fixes

Fixes #434  by @simonvadee 

## Description

* script in pyrog-server
* create user in pyrog db
* configurable email/password/role (via env vars )

### DIFF
--- a/server/.env
+++ b/server/.env
@@ -7,7 +7,5 @@ FHIR_API_URL=https://demo.arkhn.com/api
 USER_INFO_URL=http://localhost:4444/userinfo
 OAUTH2_CLIENT_ID=pyrog-server
 OAUTH2_TOKEN_URL=http://127.0.0.1:4444/oauth2/token
-SUPERUSER_EMAIL=admin@arkhn.com
-SUPERUSER_PASSWORD=superuser_password
 # Uncomment the following line if you wish to see prisma debug logs
 #DEBUG="*"
--- a/server/.env
+++ b/server/.env
@@ -7,5 +7,7 @@ FHIR_API_URL=https://demo.arkhn.com/api
 USER_INFO_URL=http://localhost:4444/userinfo
 OAUTH2_CLIENT_ID=pyrog-server
 OAUTH2_TOKEN_URL=http://127.0.0.1:4444/oauth2/token
+SUPERUSER_EMAIL=admin@arkhn.com
+SUPERUSER_PASSWORD=superuser_password
 # Uncomment the following line if you wish to see prisma debug logs
 #DEBUG="*"
--- a/server/.env.docker
+++ b/server/.env.docker
@@ -4,3 +4,6 @@ APP_SECRET=deadbeefdeadbeefdeadbeefdeadbeef
 FHIR_API_URL=http://reverse-proxy/api
 REDIS_URL=redis://redis:6379
 ENV=test
+SUPERUSER_EMAIL=admin@arkhn.com
+SUPERUSER_PASSWORD=password
+IDENTITY_PROVIDER_URL=https://staging.arkhn.com/identity-provider
--- a/server/.env.docker
+++ b/server/.env.docker
@@ -4,6 +4,3 @@ APP_SECRET=deadbeefdeadbeefdeadbeefdeadbeef
 FHIR_API_URL=http://reverse-proxy/api
 REDIS_URL=redis://redis:6379
 ENV=test
-SUPERUSER_EMAIL=admin@arkhn.com
-SUPERUSER_PASSWORD=password
-IDENTITY_PROVIDER_URL=https://staging.arkhn.com/identity-provider
--- a/server/docker-entrypoint.sh
+++ b/server/docker-entrypoint.sh
@@ -11,5 +11,8 @@ yarn wait:db
 # apply migrations
 yarn migrate:up
 
+# create superuser
+yarn seed:superuser
+
 # run the app
 yarn start
--- a/server/docker-entrypoint.sh
+++ b/server/docker-entrypoint.sh
@@ -11,8 +11,5 @@ yarn wait:db
 # apply migrations
 yarn migrate:up
 
-# create superuser
-yarn seed:superuser
-
 # run the app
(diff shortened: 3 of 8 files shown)


server/scripts/createSuperUser.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+prismaClient
+  .$connect()
+  .then(() => main())
+  .then(() => prismaClient.$disconnect())
+  .then(() => process.exit(0))
+  .catch(err => {
+    console.error(err)
+    prismaClient.$disconnect().then(() => process.exit(+!SUPERUSER_PASSWORD))
```
simonvadee, line 26: why not just process.exit(1) ?

Jasopaum (commented):
We also wanted to signup the admin with the identity-provider, do you think you could that?
The idea would be to try to signup with IdP and then create the user with the same email in Pyrog. And if the signup fails, we shouldn't insert the user in Pyrog

server/.env
```diff
@@ -7,5 +7,7 @@ FHIR_API_URL=https://demo.arkhn.com/api
 USER_INFO_URL=http://localhost:4444/userinfo
 OAUTH2_CLIENT_ID=pyrog-server
 OAUTH2_TOKEN_URL=http://127.0.0.1:4444/oauth2/token
+SUPERUSER_EMAIL=admin@arkhn.com
```
simonvadee, line 10: it may be confusing to add these lines to the .env (one may think that it is required to run pyrog-server). I would remove these two lines and document how to use the seed:superuser command in the README, what do you think ?

server/scripts/createSuperUser.ts
```diff
@@ -0,0 +1,42 @@
...
+    name: 'admin',
+    password: SUPERUSER_PASSWORD as string,
+    email: SUPERUSER_EMAIL || 'admin@arkhn.com',
+  }
+  let form = new FormData()
+  form.append('name', user.name)
+  form.append('password', user.password)
+  form.append('email', user.email)
+  const res = await axios.post(`${IDENTITY_PROVIDER_URL}/signup`, form)
+  if (res.status === 500)
```
Jasopaum, line 24: Not sure what's the best but wouldn't checking if `res.status !== 200` be "safer"?

server/scripts/createSuperUser.ts
```diff
@@ -0,0 +1,43 @@
...
+  const user = {
+    name: 'admin',
+    password: SUPERUSER_PASSWORD as string,
+    email: SUPERUSER_EMAIL || 'admin@arkhn.com',
+  }
+  let form = new FormData()
+  form.append('name', user.name)
+  form.append('password', user.password)
+  form.append('email', user.email)
+  await axios.post(`${IDENTITY_PROVIDER_URL}/signup`, form, {
```
Jasopaum, line 24: Does this raise an Exception if the signup fails?

Jasopaum (approved):

server/package.json
```diff
@@ -26,6 +26,7 @@
   "dependencies": {
     "@arkhn/fhir.ts": "^3.4.3",
     "@nexus/schema": "^0.14.0",
+    "@types/form-data": "^2.5.0",
```
simonvadee, line 29: this goes to devDependencies

server/.env.docker
```diff
@@ -3,4 +3,7 @@ POSTGRES_URL=postgresql://prisma:prisma@postgres:5432/prisma?schema=pyrog
 APP_SECRET=deadbeefdeadbeefdeadbeefdeadbeef
 FHIR_API_URL=http://reverse-proxy/api
 REDIS_URL=redis://redis:6379
-ENV=test
\ No newline at end of file
+ENV=test
+SUPERUSER_EMAIL=admin@arkhn.com
+SUPERUSER_PASSWORD=password
+IDENTITY_PROVIDER_URL=https://staging.arkhn.com/identity-provider
```
simonvadee, line 9: not sure we should leave staging URL in docker env vile

server/scripts/createSuperUser.ts
simonvadee: :+1: it should not raise an exception (the user may already exist in idp and it's fine)

server/docker-entrypoint.sh
```diff
@@ -11,5 +11,8 @@ yarn wait:db
 # apply migrations
 yarn migrate:up
 
+# create superuser
+yarn seed:superuser
```
simonvadee, line 15: I don't think we should add that to the entrypoint since it will force us to provide the superuser password in the environment all the time (event if the superuser exists in pyrog db)

server/scripts/createSuperUser.ts
```diff
@@ -1,16 +1,33 @@
...
-const { SUPERUSER_EMAIL, SUPERUSER_PASSWORD } = process.env
+const {
+  SUPERUSER_EMAIL,
+  SUPERUSER_PASSWORD,
+  IDENTITY_PROVIDER_URL,
+} = process.env
 
 const main = async () => {
   if (!SUPERUSER_PASSWORD) throw Error('superuser password is required')
+  const user = {
```
simonvadee, line 15: maybe we should add a check whether the user already exist in pyrog and skip everything if that's the case, what do you think ?

server/scripts/createSuperUser.ts
```diff
@@ -12,6 +12,10 @@ const {
 
 const main = async () => {
   if (!SUPERUSER_PASSWORD) throw Error('superuser password is required')
+  const res = await prismaClient.user.count({
+    where: { name: 'admin', email: SUPERUSER_EMAIL },
+  })
+  if (res) return
```
simonvadee, line 18: please add a small log like console.log(`user ${SUPERUSER_EMAIL} already exists, skipping...`)

simonvadee (approved):

PR #442, opened 2020-11-09
disable ACL in test environment

simonvadee:
In the context of E2E tests, there is no user in the pyrog DB, hence we cannot create an ACL row to connect to the user creating the source.

I'm wondering if I should implement the whole authentication workflow in E2E tests to go around this issue, what's your opinion ?

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -20,12 +20,12 @@ export const getUser = async (
 ): Promise<User | null> => {
   if (!IN_PROD) {
     // insert a fake admin user for tests
-    await prisma.user.upsert({
+    const user = prisma.user.upsert({
       where: { id: TEST_ADMIN_USER.id },
       create: TEST_ADMIN_USER,
       update: TEST_ADMIN_USER,
     })
-    return TEST_ADMIN_USER
+    return user
   }
 
   const authorization = request.get('Authorization')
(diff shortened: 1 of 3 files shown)


simonvadee:
actually i'll update the way we mock the user (see getUser in utils.ts) in a test environment rather than doing this.

server/src/utils.ts
```diff
@@ -11,15 +11,22 @@ import {
...
-    // return a fake user for tests
-    return { id: 'admin', name: 'admin', email: 'admin', role: 'ADMIN' } as User
+  if (!IN_PROD) {
+    // insert a fake admin user for tests
+    await prisma.user.upsert({
+      where: { id: TEST_ADMIN_USER.id },
+      create: TEST_ADMIN_USER,
+      update: TEST_ADMIN_USER,
+    })
+    return TEST_ADMIN_USER
```
Jasopaum, line 28: It's equivalent but I would have written
```
const user = await prisma.user.upsert({ ... })
return user
```
What do you think?

Jasopaum (approved):
Looks better than before 💪

server/src/utils.ts
simonvadee: 👍 done

Jasopaum: Noice 👍

PR #453, opened 2020-11-23
Add relation when importing condition

### LINKED ISSUE #448: The condition relation is lost during mapping import
## Description

When we import a mapping on pyrog with condition, the relation is not taken into account.

## Reproduction

1. Export a mapping containing the condition which have a relation other than "=="
2. Import the source

## Expectation

Keep the relation

## Screenshots

Before:
image

After:
image

## Environment

Importing from staging to dev or from staging to staging

## Additional context

The relation is correctly stored in the mapping
image

Jasopaum:
## Fixes
Fixes #448 by @nriss 

## Description
See issue


simonvadee (approved):

nriss:
Thanks 🙏

PR #454, opened 2020-11-23
Don't override authorization header if already present

Jasopaum:
## Fixes
A error was returned by Hydra when trying to revoke a token because the auth header was being overidden by the interceptor?


simonvadee (approved):
trusting you on this

PR #455, opened 2020-11-24
[auth] upsert superuser and skip IDP

simonvadee:


Jasopaum (approved):
If the signup fails, we don't upsert in pyrog, right?

simonvadee:
no we don't since an exception is raised, the program exits

PR #465, opened 2020-12-03
Small auth improvements

### LINKED ISSUE #429: add an option to bypass authentication in pyrog-client
when running pyrog (client and/or server) locally, we don't want to setup hydra, therefore we need an option to bypass/disable authentication.

we already have this in pyrog-server (ENV="test") but it could use a small refacto (use a AUTH_DISABLED env var like fhir-api ?)

### LINKED ISSUE #462: Raise error message when email does not exist
## Problem
The wheel keeps spinning when the connection failed if the email is null.

## Description
Error from server: 
```
Invalid options provided to ApolloServer: 
Invalid `prisma.user.upsert()` invocation in
/srv/dist/utils.js:114:55
{
  where: {
    email: null
           ~~~~
  },
  create: {
    email: null,
           ~~~~
    name: 'John'
  },
  update: {
    name: 'John'
  }
}
Argument email for where.email must not be null. Please use undefined instead.
Argument email: Got invalid value null on prisma.upsertOneUser. Provided null, expected String.

```

Thanks 🥨

Jasopaum:
Fixes #429
Fixes #462


client/src/constants/index.ts
```diff
@@ -3,6 +3,11 @@ export const REFRESH_TOKEN_STORAGE_KEY = 'ARKHN_REFRESH_TOKEN';
 export const ID_TOKEN_STORAGE_KEY = 'ARKHN_ID_TOKEN';
 export const TOKEN_DATA_STORAGE_KEY = 'ARKHN_TOKEN_DATA';
 export const STATE_STORAGE_KEY = 'ARKHN_AUTH_STATE';
+export const AUTH_DISABLED =
+  process.env.REACT_APP_AUTH_DISABLED &&
```
simonvadee, line 7: do you need the `process.env.REACT_APP_AUTH_DISABLED &&` if you use the question mark syntax ?

server/src/server.ts
```diff
@@ -39,7 +39,7 @@ if (IN_PROD) {
     async error => {
       const originalRequest = error.config
 
-      if (IN_PROD && error.response.status === 401 && !originalRequest._retry) {
+      if (error.response.status === 401 && !originalRequest._retry) {
```
simonvadee, line 42: why removing this ?

server/src/permissions/index.ts
```diff
@@ -17,24 +18,28 @@ export const authorizationError = {
 // We want to skip authorization checks during integration tests
 const skipAuthorizations = ENV === 'test'
 
+const checkUser = (user: User | undefined): Error | undefined => {
+  if (!user) {
+    return new Error(
+      `${authenticationError.code}: ${authenticationError.message}`,
+    )
+  }
+  if (!user.email) {
```
simonvadee, line 27: I don't understand when this happen ? how can a user.email be null if user is not ?

client/src/constants/index.ts
Jasopaum: Hum, right, it'll be better!

server/src/server.ts
Jasopaum: We're already in a `if (IN_PROD)`...

server/src/permissions/index.ts
Jasopaum: This happens when the idp doesn't provide an email for the user

client/src/constants/index.ts
Jasopaum: Oh, the argument of `includes` can't be undefined... I can still remove the question mark though

simonvadee: right, it's fine :+1:

server/src/permissions/index.ts
simonvadee: right

simonvadee (approved):

PR #474, opened 2020-12-24
feat: handle batch cancel in River view

### LINKED ISSUE #473: Add the "cancel batch" button

tevariou:
## Fixes

Fixes #473 by @simonvadee


client/src/components/river/index.tsx
```diff
@@ -155,11 +186,36 @@ const FhirRiverView = (): React.ReactElement => {
...
+          )}
+          inputItem={
+            selectedBatch
+              ? `${selectedBatch.id} ${selectedBatch.timestamp}`
+              : 'Select batch to cancel'
+          }
+          onChange={(item: string): void =>
+            setSelectedBatch({
+              id: item.split(' ')[0],
+              timestamp: item.split(' ')[1]
```
simonvadee, line 206: lol this is pretty hacky to say the least. I'm ok to leave it like this since this pyrog won't last very long, but it would have been cleaner to use only the timestamp in this select, or write a BatchSelect.

simonvadee (commented):
since the state is persisted, how will the batch disappear from the list if it completes normally without being canceled ? We don't (yet) have a way to list the current batches from river-api

client/src/components/river/index.tsx
Jasopaum: The client will survive, right? 👀

Jasopaum (commented):
I don't how more complex it would be but wouldn't it be nice if the information was fetched from river?
Maybe I missed something but does the batch stays there if it terminates by itself?

My UI two cents: I would have displayed all the current batches instead of hiding them in a StringSelect but it's ok, we can do a second round on this

tevariou:
@Jasopaum we can add a route to River-API. The batch list can be retrieved from Redis.

client/src/services/batchList/actions.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { RIVER_URL } from '../../constants';
+
+const listBatch = (): IAction => {
+  return async (dispatch: any) => {
+    try {
+      const { data } = await axios.get(`${RIVER_URL}/batch`);
+      dispatch({
+        type: 'LIST_BATCH_SUCCESS',
+        payload: data
+      });
```
Jasopaum, line 13: @BPierrick, is it a good way to handle redux and axios calls?

tevariou: One proper modern way would have been to use RTK: first a slice which include the reducer and its actions and here an async thunk (https://redux-toolkit.js.org/api/createAsyncThunk) handling the request and its side effects. Usually  we also need to create a wrapper around axios instead of calling it directly and use interceptors if needed for each API. 

But since we don't use redux toolkit in this project and there's no pre-established architecture of the redux store for handling side effects I chose to not do any refactoring especially since it will soon become legacy code.

client/src/components/river/index.tsx
```diff
@@ -90,7 +105,7 @@ const FhirRiverView = (): React.ReactElement => {
     setRunning(true);
 
     try {
-      const response = await axios.post(
+      await axios.post(
```
simonvadee: why is the call to GET /batch is in an action but not this call ? because it does not impact the application state ?

client/src/graphql/fragments/entireCredential.graphql
```diff
@@ -6,9 +6,6 @@ fragment entireCredential on Credential {
   model
   login
   decryptedPassword
-  owners {
-    id
-    name
-    schema
-  }
```
simonvadee: ???

client/src/components/river/index.tsx
```diff
@@ -138,7 +189,7 @@ const FhirRiverView = (): React.ReactElement => {
         <p>
           Choose all the resources you want to process with the ETL. If you
           select none of them, all we be processed. Note that you need to select
-          a source before.
+          a source before
```
Jasopaum, line 192: Could you remove the `If you select none of them, all we be processed.`, please 🙏

client/src/components/river/index.tsx
```diff
@@ -115,6 +147,25 @@ const FhirRiverView = (): React.ReactElement => {
     setRunning(false);
   };
 
+  const onClickCancelBatch = async (
+    e: React.MouseEvent<HTMLElement, MouseEvent>
+  ): Promise<void> => {
+    e.preventDefault();
```
Jasopaum, line 153: Why is there a preventDefault?

client/src/components/river/index.tsx
```diff
@@ -35,17 +38,44 @@ const qSourcesAndResources = loader(
...
   const [selectedResources, setSelectedResources] = useState([] as Resource[]);
   const [running, setRunning] = useState(false);
+  const [selectedBatch, setSelectedBatch] = useState('');
 
   const { data } = useQuery(qSourcesAndResources, {
     fetchPolicy: 'no-cache'
   });
 
+  useEffect(() => {
+    dispatch(listBatch());
```
Jasopaum, line 55: This will only happen once when the component is mounted, right? Is it what we want?
Do we need something to refresh the list of batches?

client/src/components/river/index.tsx
```diff
@@ -155,11 +206,60 @@ const FhirRiverView = (): React.ReactElement => {
...
           </Button>
         </div>
+        <h1>Cancel a batch</h1>
+        <StringSelect
+          items={
+            batchList.data
+              ? Object.keys(batchList.data).map(
+                  (batchId: string) => batchList[batchId].timestamp
+                )
+              : []
```
Jasopaum, line 221: What do you think about destructing the query response in the action? I don't know what people usually do I think it would be nice to avoid this logic here and put it closer to the call

client/src/components/river/index.tsx
```diff
@@ -155,11 +206,60 @@ const FhirRiverView = (): React.ReactElement => {
...
+          }
+          onChange={(item: string): void => {
+            const batchId = Object.keys(batchList.data).find(
+              batchId => batchList[batchId].timestamp === item
+            );
+            if (batchId) setSelectedBatch(batchId);
+          }}
+        />
+        <ul>
+          {!!selectedBatch &&
```
Jasopaum, line 236: nit: are the `!!` really needed?

client/src/components/river/index.tsx
```diff
@@ -155,11 +206,60 @@ const FhirRiverView = (): React.ReactElement => {
...
+              : 'Select a batch to cancel'
+          }
+          onChange={(item: string): void => {
+            const batchId = Object.keys(batchList.data).find(
+              batchId => batchList[batchId].timestamp === item
+            );
+            if (batchId) setSelectedBatch(batchId);
+          }}
+        />
+        <ul>
```
Jasopaum, line 235: This is a list of the resources processed in the selected batch, right?
Could you add a small paragraph or find another way to make that clear to the user?

client/src/components/river/index.tsx
tevariou: The action should stay minimal https://redux.js.org/style-guide/style-guide#put-as-much-logic-as-possible-in-reducers . Also the reducer should store normalized data in the global state and handle errors as well. What I can do is add a selector.

client/src/components/river/index.tsx
Jasopaum: Oh right, maybe it's the selector's job to do that. What do you think?

client/src/components/river/index.tsx
Jasopaum: Oh right, maybe it's the selector's job to do that. What do you think?

client/src/components/river/index.tsx
Jasopaum: Agreed!
I'm just worried that batches that have terminated will appear but it's not that severe

client/src/components/river/index.tsx
tevariou: I think that was why it has been done like this. It may not be worth it to store the result of those queries in the global state since it is not used anywhere else.

tevariou:
@simonvadee @Jasopaum can we merge this?

client/src/services/batchList/reducer.tsx
```diff
@@ -0,0 +1,35 @@
...
+const initialState: BatchResponse = {
+  data: {},
+  error: null
+};
+
+const batchList = (state = initialState, action: ISimpleAction): any => {
+  switch (action.type) {
+    case 'LIST_BATCH_SUCCESS': {
+      return {
+        data: action.payload
```
Jasopaum, line 12: Why is this ternary needed?

simonvadee (approved):
lgtm

client/src/services/batchList/reducer.tsx
tevariou: The api returns null when there's no result (instead of an empty array). Maybe we should change that in another pr.

tevariou:
@simonvadee I have tested the frontend locally using the APIs on staging

PR #482, opened 2021-01-14
Improve UI

### LINKED ISSUE #480: User feedback of brand new Pyrog
## Essential Feedback
- pre-fill join tables when the SQL input table, or filter table differs from the reference table. 
- the orange dots (= filled attributes) stay filled after deleting attributes

## Nice to have
- move 'remove input button' (X) above the SQL input or condition, to avoid component overlapping 
- when selecting an attribute, the first click 'add an input group' is superfluous. We can discuss how we can avoid this 

## Very nice to have
- undo : revert a click that removed the condition
- suggest sql input based on last sql input

Jasopaum:
## Fixes
Fixes partially #480


client/src/components/mapping/InputColumn/index.tsx
```diff
@@ -68,6 +78,12 @@ const InputColumn = ({ input }: Props) => {
         inputId: input.id
       }
     });
+    const updatedGroup = attributeRules.inputGroups.find(
+      group => group.id === input.inputGroupId
+    );
+    updatedGroup!.inputs = updatedGroup!.inputs.filter(i => i.id !== input.id);
```
simonvadee, line 84: I had an idea to replace this (maybe), can we use the return value of the mutation deleteInput above ? this was we retrieve directly the up-to-date attribute mapping and no need to change it manually

client/src/components/mapping/InputGroup/index.tsx
```diff
@@ -78,21 +80,37 @@ const InputGroup = ({ inputGroup, key }: Props) => {
...
+  const onDeleteInputGroup = () => {
+    deleteInputGroup({
+      variables: {
+        attributeId: inputGroup.attributeId,
+        inputGroupId: inputGroup.id
+      }
+    });
+    attribute.inputGroups = attribute.inputGroups.filter(
+      g => g.id !== inputGroup.id
+    );
```
simonvadee, line 92: same as above

client/src/components/mapping/InputStatic/index.tsx
```diff
@@ -28,10 +29,12 @@ const mUpdateStaticInput = loader(
 const mDeleteInput = loader('src/graphql/mutations/deleteInput.graphql');
 
 interface Props {
+  attributeRules: IAttribute;
   input: IInput;
 }
 
-const InputStatic = ({ input }: Props): React.ReactElement => {
+const InputStatic = ({ attributeRules, input }: Props): React.ReactElement => {
```
simonvadee, line 36: attributeRules -> attribute

client/src/components/mapping/InputStatic/index.tsx
```diff
@@ -72,6 +75,12 @@ const InputStatic = ({ input }: Props): React.ReactElement => {
         inputId: input.id
       }
     });
+    const updatedGroup = attributeRules.inputGroups.find(
+      group => group.id === input.inputGroupId
+    );
+    updatedGroup!.inputs = updatedGroup!.inputs.filter(i => i.id !== input.id);
+
```
simonvadee, line 82: use mutation return value ?

simonvadee (approved):

PR #499, opened 2021-02-25
Fix filters and change pagai url

### LINKED ISSUE #497: Filters not saved on Ressource declaration
## Description
When a filter on the Ressource table is added on the Ressource (in the window where we give information about the Primary Key), the filter is not saved (part in hook) and the Ressource table does not appear. 

[source / table / attribut] = xxx. 

**Erreur in Network :** {"error":"'NoneType' object is not subscriptable"}

## Reproduction
1. Create a ressource
2. Select a Primary Key from the Ressource table
3. Add a filter from the Ressource table 
4. Save the Ressource 
5. See error.

## Expectation
We should see the Ressource table, without errors.as one the old Pyrog.  

## Environment
Staging, Toulouse

### LINKED ISSUE #498: Filter from another table in the creation of a new ressource changes the join
## Description
When a filter from another table is added on the ressource (in the window to give the information about the Primary Key),we have to add a join. After the addition of all those information and the saving of the ressource, the error appears. 

Erreur in Network in the case of the _example_ : {"error":"Don't know how to join to public.patients; please use an ON clause to more clearly establish the left side of this join"}

If we refresh the page, the join that we did has changed. The second table become the first table (see example) : 

_Example_ : Creation of a Patient Ressource based on the patients table. 
> Ressource Patient : Primary Key > public.patients.row_id
> Filter on admissions table : public.admissions.language = fr 
> Join on : public.**patients**.subject_id = public.**admissions**.subject_id. 
> 
> After refresh the page there is : 
> Ressource Patient : Primary Key > public.patients.row_id
> Filter on admissions table : public.admissions.language = fr 
> Join on : public.**admissions**.subject_id = public.**patients**.subject_id

## Reproduction
1. Create a ressource
2. Select a Primary Key from a ressource table
3. Add a filter from another table and do a join between ressource table and filter table. 
4. Save the ressource 
5. Refresh the page
6. See error.

## Expectation
We should see the ressource table, without errors. And the join should be in the right order

## Environment
Staging, Toulouse

Jasopaum:
Fixes #497
Fixes #498


simonvadee (approved):
💯

PR #507, opened 2021-04-12
Add IN relation to conditions

Jasopaum:
Quick addition for urgent need
Did I miss something?


simonvadee (approved):
I think you're good, it deserves to be deployed on dev before merging though.
also, can you create an issue on the new pyrog to propagate this change pls ?

PR #525, opened 2021-06-08
Add the possibility to select an update frequency when creating batches

Jasopaum:


client/src/components/river/index.tsx
```diff
@@ -109,17 +113,26 @@ const FhirRiverView = (): React.ReactElement => {
 
   const handleClickCreateBatch = async (): Promise<void> => {
     setRunning(true);
+    const runOnce = selectedFrequency === "run once"
```
simonvadee, line 116: naming confused me at first, I suggest 
```suggestion
    const isRecurring = selectedFrequency !== "run once"
```

simonvadee: I think it will make the rest of this function clearer

client/src/components/river/index.tsx
```diff
@@ -109,17 +113,26 @@ const FhirRiverView = (): React.ReactElement => {
...
+    const route =  runOnce? "api/batch/" : "api/update-batch/";
```
simonvadee, line 117: very nit: I would have put this ternary directly in the template string when calling axios

client/src/components/selects/stringSelect.tsx
```diff
@@ -54,13 +55,17 @@ export default class StringSelect extends React.Component<ISelectProps, any> {
...
       inputItem,
       intent,
-      items,
       loading,
       onChange,
       filterable,
       popoverProps
     } = this.props;
 
+    const doSort = (this.props.doSort === undefined || this.props.doSort === true);
```
simonvadee, line 64: can't you have default props value instead ?

simonvadee (changes requested):

simonvadee (commented):
do we want to merge this one or leave it as a separate branch since we're not going to merge [this one](https://github.com/arkhn/fhir-river/pull/443) ?

Jasopaum:
We don't use pyrog anymore

PR #533, opened 2021-07-12
feat(client): add switch for service name connection

tevariou:
## Fixes

Fixes https://github.com/arkhn/fhir-river/issues/499

Screenshot 2021-07-12 at 19 11 34
Screenshot 2021-07-12 at 18 25 18

### DIFF
--- a/client/src/components/newSource/index.tsx
+++ b/client/src/components/newSource/index.tsx
@@ -4,6 +4,7 @@ import {
   FormGroup,
   InputGroup,
   HTMLSelect,
+  Switch,
   Icon
 } from '@blueprintjs/core';
 import axios from 'axios';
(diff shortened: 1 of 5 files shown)


simonvadee (approved):
lgtm

client/src/components/newSource/index.tsx
```diff
@@ -375,6 +378,12 @@ const NewSourceView = (): React.ReactElement => {
                 }}
                 placeholder={'Database name'}
               />
+              <Switch
```
Jasopaum, line 381: Doesn't look pretty, right? 😄

Jasopaum (commented):
A bit schlag but I know we don't really want to change the data model
We should add as issue to do this in the new pyrog, right?

